PR #91, opened 2021-04-16
Migrate ChargeRepository from green-energy-hub-archived

BjarkeMeier:
## Description
Migrated IChargeRepository.cs and its implementation and unit tests from green-energy-hub-archived.

It has been decided to neither refactor nor redesign the code but basically just migrate it to geh-charges. Thus this PR provides a handle for persisting change of charge transactions. This will be done in an upcoming PR.

## References

* https://dev.azure.com/energinet/Datahub/_workitems/edit/118414

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeTypePeriod.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeTypePeriod.cs
@@ -18,8 +18,13 @@ namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
 {
     public class ChargeTypePeriod
     {
+        public ChargeTypePeriod(List<Point> points)
+        {
+            Points = points;
+        }
+
         public string? Resolution { get; set; }
 
-        public List<Point>? Points { get; }
+        public List<Point> Points { get; }
     }
 }
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/Context/Model/ResolutionType.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/Context/Model/ResolutionType.cs
@@ -12,22 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System.Text.Json.Serialization;
-using NodaTime;
-
-namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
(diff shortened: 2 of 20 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeTypePeriod.cs
```diff
@@ -18,8 +18,13 @@ namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
...
     public class ChargeTypePeriod
     {
+        public ChargeTypePeriod(List<Point> points)
+        {
+            Points = points;
+        }
+
         public string? Resolution { get; set; }
 
-        public List<Point>? Points { get; }
```
lasrinnil, line 23: Perhaps we could instantiate the collection directly from the constructor and instead have a method for adding the points to the collection externally? This way we wouldn't have to null the Points collection before using it.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/Context/Model/ResolutionType.cs
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace GreenEnergyHub.Charges.Infrastructure.Context.Model
+{
+    public class ResolutionType
+    {
+        public int Id { get; set; }
+
+        public string? Name { get; set; }
```
lasrinnil, line 21: Question: Is a ResolutionType without a name valid?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeTypePeriod.cs
BjarkeMeier: That's actually exactly what I did ;-)

lasrinnil (approved):
Apporved !

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/Context/Model/ResolutionType.cs
BjarkeMeier: I don't think it would be valid.
Since we went with C# 8.0 and Nullable Reference Types all these nullable problems - that already existed now just have become explicit. We still need to find our way regarding how to get rid of nullable where they are now allowed while still keeping clean code (readable and maintainable).

Mech0z (approved):

PR #101, opened 2021-04-20
Improve AddGreenEnergyHub performance

mhejle:
## Description
The algoritm for finding and registering IHubMessages was O(n^2) or worse.
For every found type, the algoritm had to check the ever growing list of previously found types.

Given that the algorithm checks an unknown number of assemblies with an unknown number of
IHubMessage implementations - this can quickly get out of hand.

Replace the code with an algorithm that runs in O(n).
The code now only handles each type once, as we rely on the dictionary to filter out duplicates
and enforcy priority based on assembly order.

--
Note, even with the added unittests the postcondition that assemblies are prioritized in the given order, is not checked.
In order to test this, one would have to add IHubMessage implementations in multiple assemblies, which i am not confident enough to do.

The same goes for the fact that framework assemblies are checked for implementations.
If there were any IMessageHub implementations in the framework - this test would verify the postcondition

```
        [Fact]
        public void HandlerExtension_ShouldRegisterIHubMessagesFromFrameworkAssemblyAutomatically()
        {
            var serviceCollection = new ServiceCollection();
            serviceCollection.AddGreenEnergyHub(); // adding none, relying on framework assembly being added automatically
            var messageRegistrations = serviceCollection.Where(_ => _.ServiceType == typeof(MessageRegistration));

            Assert.Single(messageRegistrations);
        }
```

I have tested both variations manually by adding various IHubMessage implementations to the code and debugging the test, These changes i rolled back as to not polute the GreenEnergyHub.Messaging project.

--
Comment: 
The flow in AddGreenEnergyHub seems frail as multiple cases depend on the list of assemblies being constructed.
- services.AddMediatR(assemblies);
- services.DiscoverValidation(assemblies);
- search for IHubMessages

Where atleast the IHubMessage functionality depends on a certain order of assemblies.
An idea for a future improvement might be to add a more explicit handling of the assembly list - with any preconditions checked and mentioned.

On another note, 
DiscoverValidation extracts all types from the assembly list - just like the IHubMessage functionality does.
It might be worth considering, making a common utility to isolate this responsibility.


BjarkeMeier:
@mhejle , thank you very much for your contribution. It looks solid. We'll have an internal talk about how to integrate your suggestion. You're addressing an issue in "shared code" which for now is simply copied around to a number of domains. So we want to decide whether to update our copy or perhaps move it to a shared repo and possibly distribute as NuGet package or other.

We'll get back to you as soon as possible.

bemwen (approved):
@mhejle: As mentioned by @BjarkeMeier this is part of some shared code which does not live in the charge domain.

However, since the placement of the shared code is not currently decided we will be doing the code review and applying it here in the charges domain.

Later, the commit will then be moved and merged into its proper location´, once that is known.

Again, thanks you for you contribution which is looking really good.

mhejle:
> @mhejle: As mentioned by @BjarkeMeier this is part of some shared code which does not live in the charge domain.
> 
> However, since the placement of the shared code is not currently decided we will be doing the code review and applying it here in the charges domain.
> 
> Later, the commit will then be moved and merged into its proper location´, once that is known.
> 
> Again, thanks you for you contribution which is looking really good.

Thanks for the update - i have spotted a few minor things i would like to change before hand (nothing functional - mostly bad wording)
Will post force an update of the PR shortly

source/messaging/source/GreenEnergyHub.Messaging.Integration.ServiceCollection/HandlerExtensions.cs
```diff
@@ -82,5 +68,34 @@ public static void AddGreenEnergyHub(this IServiceCollection services, params Sy
...
+        /// <param name="assemblies">A sequence of assemblies, in which to search for implementations of <see cref="interfaceToCheckFor"/>.</param>
+        private static List<Type> FindImplementations(Type interfaceToCheckFor, System.Reflection.Assembly[] assemblies)
+        {
+            var hubMessageNameToTypeDictionary = new Dictionary<string, Type>();
+
+            // Check the assemblies in reverse order to satisfy the post condition that the first encounter of a given name wins.
+            // this works since '[type.Name] = type' adds/updates the type value stored, overriding values from less significant assemblies.
+            foreach (var assemblyToCheck in assemblies.Reverse())
+            {
+                var messageTypes = assemblyToCheck.GetTypes()
```
Sondergaard, line 89: The invocation of `.Reverse()`can be avoided by looping the array from last index to first

```suggestion
            for (var i = assemblies.Length - 1; i >= 0; i--)
            {
             var messageTypes = assemblies[i].GetTypes()
```

Sondergaard (changes requested):
Hi thank you for the PR.

I see some hiccups from the compiler with the xmldocs.

Please use [<paramref />](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/paramref) when referencing method parameters - I believe that should do the trick.

NIT: I have added a single suggestion in regards to looping the assemblies array.

source/messaging/source/GreenEnergyHub.Messaging.Integration.ServiceCollection/HandlerExtensions.cs
mhejle: Fair point, since i did argument from a performance perspective i should have done as you suggested.
My reasoning for not doing it, was readability - and given that a reverse is O(n) i did not mind the performance hit.

All taken into account - i think your suggestion is a better option

mhejle:
> Hi thank you for the PR.
> 
> I see some hiccups from the compiler with the xmldocs.
> 
> Please use when referencing method parameters - I believe that should do the trick.
> 
> NIT: I have added a single suggestion in regards to looping the assemblies array.

I assume your comment was meant to say : "use `paramref` when ..." i will update the PR accordingly

mhejle:
@Sondergaard I have updated the PR with the requested changes.
Thank you for the feedback

Sondergaard (approved):
LGTM

PR #107, opened 2021-04-23
PreDeploy and PostDeploy scripts included per default

HenrikSommer:
## Description

PreDeploy and PostDeploy scripts are included per default, but excluded using parameters in db migration pipeline, enabling us to migrate database without enabling CDC.


bemwen (approved):

BjarkeMeier (approved):
LGTM. See my single minor comment.

PR #149, opened 2021-04-27
Rejection flow (with exceptionhandling)

lasrinnil:
## Description

This PR will create handling of the rejection flow. 

If exceptions are thrown, they now also will be handled by the rejection flow.

I have tested the rejection flow on my local sandbox.
image

And verified the Accepted flow still works
image

## References

* #0000

PR #161, opened 2021-04-28
Unify inputvalidation with businessvalidation

HenrikSommer:
As outcome of this PR inputvalidation of incoming charge is now performed in the same manner as businessvalidation, making mapping of inputvalidation validationresults obsolete.

Inputvalidation is now under test.

This PR builds upon https://github.com/Energinet-DataHub/geh-charges/pull/149


lasrinnil:
Perhaps we ought to redo the merge together when we have time @HenrikSommer. There are files and changes in your PR that are  leftovers from where my branch left off.

Issue #116, opened 2021-04-26
ChargeCommand - Update NamingScheme as per guidebook


PerLysemose:
Comment from Michael:
Denne opgave skal dække:
Navngivning af chargecommand
Et blik på nullables

PerTHenriksen:
Link to engelsk navngivning på forskrifter: https://en.energinet.dk/Electricity/Rules-and-Regulations/Archive-Market-Regulations

prtandrup:
AC#5: Postman Scripts were updated to reflect the new naming/structure of the ChargeCommand.
![image](https://user-images.githubusercontent.com/71757561/117780209-b6884600-b23f-11eb-9c50-70e280ae8317.png)
Example for Create Fee: 
`
{
    "Document": {
        "Id": "DocId{{$isoTimestamp}}",
        "CreatedDateTime": "{{$isoTimestamp}}",
        "IndustryClassification": 23,
        "DocumentType": 10,
        "Sender": {
            "Id": 37,
            "MRid": "8100000000030",
            "Name": "Grid Operator 3",
            "Role": 0
        },
        "Recipient": {
            "Id": 1,
            "MRid": "5790001330552",
            "Name": "Hub",
            "Role": 0
        }
    },
    "ChargeOperation": {
        "Id": "OpId{{$isoTimestamp}}",
        "EndDateTime": null,
        "Status": 2,
        "LastUpdatedBy": "LastUpdatedByPostman",
        "BusinessReasonCode": 18,
        "ChargeId": "ChargeId01",
        "ChargeName": "Charge Fee Name",
        "Type": 1,
        "ChargeOwner": "8100000000030",
        "Resolution": "P1D",
        "TaxIndicator": false,
        "VatClassification": 1,
        "ChargeDescription": "The charge description",
        "StartDateTime": "2021-07-31T22:00:00Z",
        "Period": {
            "Points": [
                {
                    "Position": 1,
                    "Price": 100.001,
                    "Time": "2021-07-31T22:00:00Z"
                }
            ]
        }
    }
}
`

prtandrup:
AC#4: Both PO and SME (Irene) are OK with the new naming convention. The updated dictionary can be seen on this branch [here](https://github.com/Energinet-DataHub/green-energy-hub/blob/UpdateChargesDictionary/docs/dictionary-and-concepts/dictionary-charges.md). It will be converted to a PR later today.

PR #182, opened 2021-05-03
Charge command refactor

lasrinnil:
## Description

This PR will refactor our ChargeCommand model to no longer so many null able properties. Along with checking when receiving a message that the ChargeCommand within's properties are not null where they are required by business logic to not be. 

We have tested our happy flow with serializing and de-serializing the refactored ChargeCommand.

image

## References

* #116


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeCommand.cs
```diff
@@ -21,32 +21,32 @@ namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
 {
     public class ChargeCommand : CommandBase
     {
-        public MarketDocument? MarketDocument { get; set; }
+#pragma warning disable 8618
```
BjarkeMeier, line 24: I'd like a comment about what 8618 is and why you use it because that's important and not evident from the code.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeType.cs
```diff
@@ -18,17 +18,19 @@ namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
 {
     public class ChargeType
     {
+#pragma warning disable 8618
```
BjarkeMeier, line 21: Just thoughts: Can we apply this disabling only once? E.g. using a "local" configuration file? Or by creating a project for DTOs? Or do we actually want this explicit disabling in each class?

BjarkeMeier (commented):
I assume that you have tested that the changes doesn't break the deserialization of messages received from the service bus(?)

It's nice if we can get rid of all the nullable props this way. I think the PR should also fix all the - now - superfluous null forgiving operators.

lasrinnil:
> 
> 
> I assume that you have tested that the changes doesn't break the deserialization of messages received from the service bus(?)
> 
> It's nice if we can get rid of all the nullable props this way. I think the PR should also fix all the - now - superfluous null forgiving operators.

I Would like address this specifically in a PR of its own.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeCommand.cs
lasrinnil: I have put a comment in each file ignored the CA rule now

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeType.cs
lasrinnil: We could disable it in a Globalsettings.cs file i guess, but I like that we just ignore it in the specific cases we dont want the compiler to worry about non nullable props not being instantiated

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/GlobalSuppressions.cs
```diff
@@ -0,0 +1,16 @@
+﻿//---------------------------------------------------------------------
```
BjarkeMeier: Should we remove this if it is not used?

lasrinnil: Haha, yes it is removed (should not have been checked in, DOH!)

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeCommand.cs
```diff
@@ -16,37 +16,39 @@
...
-        public MktActivityRecord? MktActivityRecord { get; set; }
+        public MktActivityRecord MktActivityRecord { get; set; }
 
         /// <summary>
         /// The kind of charge: Fee ("D02") | Subscription ("D01") | Tariff ("D03").
         /// </summary>
-        public string? Type { get; set; }
+        public string Type { get; set; }
 
         [JsonPropertyName("ChargeType_mRID")]
```
bemwen, line 37: We should try to get rid of these json properties if possible, but I assume that is out of scope of this PR?

lasrinnil: We will address in a PR which will rename the model and our Postman models at the same time me and @prtandrup

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/ChargeType.cs
```diff
@@ -13,19 +13,23 @@
 // limitations under the License.
 
 using System.Text.Json.Serialization;
+#pragma warning disable 8618
 
 namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
 {
+    // Non-nullable member is uninitialized is ignored
+    // Only properties which is allowed to be null is nullable
+    // ChargeType integrity is null checked by ChargeCommandNullChecker
```
bemwen, line 22: If I had not heard about this, I do not think I would get that comment is actually about why we are using pragma here (and the other examples). Maybe it would be easier to make the connection if they comments was nearer the pragma line.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure/Repositories/ChargeRepositoryTests.cs
```diff
@@ -199,9 +197,12 @@ private static Charge GetValidCharge()
...
             transaction.Period.AddPoints(new List<Point>
-{
-    new Point { Position = 1, PriceAmount = 1m, Time = SystemClock.Instance.GetCurrentInstant(), },
-});
+                    {
+                    new Point
+                    {
+                        Position = 1, PriceAmount = 1m, Time = SystemClock.Instance.GetCurrentInstant(),
+                    },
+                    });
```
bemwen, line 205: NIT: Maybe inline the new point segment for readability

bemwen (approved):
LGTM, only a few NIT comments

BjarkeMeier (approved):

Issue #174, opened 2021-05-03
Ensure (or document) that we have working software


PerLysemose:
-- Charge domæne mvp application insight, der er ikke vendor agnostic løsning.  -- 
inddrage SME for at høste datahub monitoreringsforventninger på den længere bane. 

Telemetry dashboard?
Monitor uptime?
Automated domain/integration tests?

Differentiate dev and test environment and gates?
Automatic notifications to the team?

ACs:
AC: dashboard der viser oversigt med events i hvilke køer. 
AC: indhold af dead letter køer. 
~|ac: time tracking, på transaktioner.~
~nice to know ac: hvor lang tid ligger events i køer.~

BjarkeMeier:
**Notes for tester**
The following ACs has been discarded due to prioritization in team Volt:

ac: time tracking, på transaktioner.
nice to know ac: hvor lang tid ligger events i køer.

prtandrup:
Tested OK on T

AC# Dashboard provides an overview of the different events in queue -> PASS
AC# Deadletter queue added -> PASS 

The stats in DeadLetter and Active messages were not verified by any test here. I suggest we take a look at this later to get a better understanding of the stats.

![image](https://user-images.githubusercontent.com/71757561/117251800-6aa35e80-ae45-11eb-975a-9b5df2b908a8.png)

PR #183, opened 2021-05-03
Add Azure portal dashboard

BjarkeMeier:
## Description

This PR adds an Azure portal dashboard for the charges domain. It is a first step towards ensuring working (running) software. It will be expanded and improved over time.

The template JSON was created by first creating the dashboard manually in the Azure Portal, then downloading it (as JSON) and then extracting the lenses-part of the file. Finally, substitution placeholders were added.

The dashboard should look like on the image. The availability test probably needs to either be fixed or removed in the future.

image

## References

[Service-bus-metrics-azure-monitor](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-metrics-azure-monitor#request-metrics)

* #174


prtandrup:
I'm puzzled by the two count options (active messages and dead letters) displaying decimals, any idea what that is?

build/infrastructure/dash-charges-template.json
```diff
@@ -0,0 +1,427 @@
...
+                          },
+                          "name": "IncomingMessages",
+                          "aggregationType": 1,
+                          "namespace": "microsoft.servicebus/namespaces",
+                          "metricVisualization": {
+                            "displayName": "Incoming Messages"
+                          }
+                        }
+                      ],
+                      "title": "Incoming Messages (sum))",
```
prtandrup, line 59: If it was the first time I saw the dashboard, this (or something similar) would be a help to me.
```suggestion
                      "title": "Incoming Messages (sum per type))",
```

BjarkeMeier: If you want that I guess you would also want (1) "sum per event type" or "sum per topic" and (2) the same addition in all 3 titles. Or...? Personally, I prefer it the way it is but I can change it.

BjarkeMeier:
> I'm puzzled by the count option for active messages displaying decimals, any idea what that is?
> And why is there an "m" after "charge-command-received" in active messages (count)

To be honest: I don't know. I didn't find the answer quickly and think I'd prefer to simply accept it for now considering the (low) priority of the story.

build/infrastructure/dash-charges-template.json
prtandrup: I can live with it as is. Just a thought :)

prtandrup:
> > I'm puzzled by the count option for active messages displaying decimals, any idea what that is?
> > And why is there an "m" after "charge-command-received" in active messages (count)
> 
> To be honest: I don't know. I didn't find the answer quickly and think I'd prefer to simply accept it for now considering the (low) priority of the story.

Agree, but at some point I would like to verify the figures displayed :)

bemwen (approved):
LGTM

prtandrup (approved):

build/infrastructure/dash-charges-template.json
BjarkeMeier: Thx for input. I'll however leave it as-is.

PR #187, opened 2021-05-05
Configure code cov

Sondergaard:
## Description

Configuration of codecov.io


BjarkeMeier (commented):
Thank you for your contribution @Sondergaard. How about giving an introduction to our team? 

The missing result seems like something is missing. Is it necessary to continuously keep adding new projects? Do we need to use other traits? Also, the distinction between unit and integration test isn't obvious to me.

Sondergaard:
> The missing result seems like something is missing. 

That is correct. There is no coverage detected since the tests are missing the category trait or the value therein does not match that of the filter. Please see below for details.

> Is it necessary to continuously keep adding new projects? 

We are utilising an online service that is adding a visualisation of our code coverage. This is already being collected and stored when the CI pipeline is invoked. When using codecov we start to actively use the collected data and illustrate how the code is covered with tests.
  
> Do we need to use other traits? 

The codecov-ci pipeline is using the filter `--filter "Category=UnitTest"` which is also being used by the other repos. The CI pipeline can be changed to another filter, but then it will diverge from the other repositories. It is my recommendation that the tests are annotated in the same way as the other repositories so that there is conformance.

The other repositories are using this nuget package [xunit.categories](https://www.nuget.org/packages/xunit.categories/) to annotate the tests.

> Also, the distinction between unit and integration test isn't obvious to me.

The codecov is configured to count all code that is found within application and domain as unit tests. Code that is found in azure functions or infrastructure is considered integration tests. This is being scoped with a concept of flags.

As for now the code coverage is only calculated on the unit part => application and domain

bemwen (approved):
LGTM, although we will need to go from the old Trait strategy to the new one to fully utilize this.

lasrinnil (approved):

PR #191, opened 2021-05-10
Service Bus subscription for integration test

HenrikSommer:
As a result of this PR we will be able to use a separate Service Bus Subscription for integration tests in the Charges domain.

For each Service Bus Topic we add a Subscription for integration test only which will enable the use of existing Service Bus Topics without interfering with data in main Subscriptions.

Integration tests for the Charges domain will be the aim of a later PR.


bemwen (approved):

build/infrastructure/sb-charges.tf
```diff
@@ -56,6 +56,14 @@ resource "azurerm_servicebus_subscription" "sbs_command_received" {
   max_delivery_count  = 1
 }
 
+resource "azurerm_servicebus_subscription" "sbs-command-integration-test" {
```
BjarkeMeier, line 59: You must use underscores.

BjarkeMeier, line 59: I'd prefer `sbs_command_received_integration_test` or perhaps invent a shorter naming convention like e.g. `sbs_command_received_it`. But I would keep the original name as a subset of the name.

BjarkeMeier (changes requested):
My comments apply to all 3 resources.

BjarkeMeier (approved):

PR #192, opened 2021-05-10
Switch test traits to xunit categories

bemwen:
## Description

The purpose of this PR is to switch our unit test from the old trait concept to using XUnit.Categories instead. The change will make the tests visible for code coverage.

The PR includes:

* Removing trait from TestHelper so that the old concept is no longer available
* Move tests using the old concept to the new one
* Added traits to tests that were missing them entirely
* Bumped Microsoft.NET.Test.Sdk so that the shared solution could again be compiled

## References

* https://github.com/Energinet-DataHub/geh-charges/pull/187 - PR adding code coverage


HenrikSommer (approved):

BjarkeMeier (approved):
LGTM

How about adding a unit test that does an assembly scanning to verify that all test classes in test project are annotated with `[UnitTest]` attribute?

I remember something about considering tests of infrastructure project as integration tests. Does that affect which attribute should be used for those tests?

BjarkeMeier (commented):
I don't think we should modify the shared stuff but rather exclude it. Shared shouldn't affect our "general" coverage as it is expected to be removed eventually.

bemwen:
Discussed over phone with the following result:

The shared code will be moved at some point which may cause of coverage to jump in either direction, but at that point we will know why it jumped.

As long as there is not a fixed percentage required by PRs it does not hurt to count shared in our coverage. If a percentage requirement is introduced later, then we will need to revisit this, possibly by removing traits in the shared code to make sure they are not counted.

Issue #124, opened 2021-04-26
Enabler: ServiceBus Communication - 3 - Implement message serialization and deserialization


PerLysemose:
**Acceptance criteria:**
AC1: An implementation of MessageSerializer supporting JSON has been implemented and shown to work through unit tests
AC2: An implementation of MessageDeserializer supporting JSON has been implemented and shown to work through unit tests

**Description:**
This story is part of moving the charge domain towards using the messaging-framework for sending messages within and between domains.

As part of this task, we need to ready implementations of MessageSerializer and MessageDeserializer that support raw JSON which we will be using internally in the charge domain.

MessageSerializer can be seen below:

```
    /// <summary>
    /// Abstract class for serializing outgoing messages.
    /// </summary>
    public abstract class MessageSerializer
    {
        /// <summary>
        /// Serialize a message
        /// </summary>
        /// <param name="message">Message to serialize</param>
        /// <param name="cancellationToken">Cancellation token for the request</param>
        /// <returns><see cref="byte"/> sequence for the message</returns>
        public abstract Task<byte[]> ToBytesAsync(IOutboundMessage message, CancellationToken cancellationToken = default);
    }
```

It will be called by the dispatcher and should make use of a mapper abstraction to map a IHubOutboundMessage to a DTO (if needed) as shown in the following diagram:

![image](https://user-images.githubusercontent.com/12642168/116399295-0d962000-a829-11eb-8a24-846ca9e51997.png)

Inspiration can be found in the current protobuf implementation found here:
https://github.com/Energinet-DataHub/geh-charges/blob/main/source/messaging/source/GreenEnergyHub.Messaging.Protobuf/ProtobufMessageSerializer.cs

Likewise, we need an implementation of MessageDeserializer that match the serializer implementation.

The abstract class looks like this:

```
    /// <summary>
    /// Abstract class for deserializing incoming messages.
    /// </summary>
    public abstract class MessageDeserializer
    {
        /// <summary>
        /// Deserialize <paramref name="data"/> to an inbound message
        /// </summary>
        /// <param name="data">Data to deserialize</param>
        /// <param name="cancellationToken">Cancellation token for the operation</param>
        /// <returns><see cref="IInboundMessage"/> extracted from <paramref name="data"/></returns>
        public abstract Task<IInboundMessage> FromBytesAsync(byte[] data, CancellationToken cancellationToken = default);
    }
```

[...]

PR #210, opened 2021-05-12
 Test that all IInboundMessages can be deserialized

BjarkeMeier:
## Description

This unit test should give us an acceptable level of confidence that all `IInboundMessage` implementations can actually be deserialized and thus received using `MessageExtractor` of the messaging framework.

## References

* #124


lasrinnil (approved):
Beautiful :)

prtandrup (approved):

Issue #195, opened 2021-05-10
Create Input validation rules for BRS-031/032/033


lasrinnil:
### Description

The goal of this story is to produce events describing which rules failed during input validation. So we in the future can handle these events accordingly.  

This story only focus on rules witch are applicable to all Create operations.

### Test

All rule implementations must be covered by unit tests.
All rejection events are expected to be tested on our T realm. 

### Rules

Rule number | Rule description | Applies to
------------ | ------------- | ------------- 
VR.488 | Only Vat classification allowed is D01 Vat or D02 NoVat 
VR.505-1 | Resolution must be P1D OR PT1H OR PT15M  | Tariff
VR.505-2 | Resolution must be P1D | Fee
VR.505-3 | Resolution must be P1M | Subscription

### Acceptance criteria

Rule number  | Acceptance criteria
------------ | -------------
VR.488 |  If Vat classification is not D01 Vat or D02 NoVat. A rejection event describing this must be found on the temporary post office service bus subscription.
VR.505-1 | If a Tariff's Resolution isn't P1D OR PT1H OR PT15M . A rejection event describing this must be found on the temporary post office service bus subscription.
VR.505-2 | If a Fee's Resolution isn't P1D. A rejection event describing this must be found on the temporary post office service bus subscription.
VR.505-3 | If a Subscription's Resolution isn't P1M. A rejection event describing this must be found on the temporary post office service bus subscription.

lasrinnil:
#### Test

VR505-2 
Tested with resolutiontype = P1M

![image](https://user-images.githubusercontent.com/6780460/118780967-e0b7b480-b88c-11eb-846a-7853236aacec.png)

PR #220, opened 2021-05-17
Validation rule VR505-2

Mech0z:
## Description
Input Validation of VR505-1

- Resolution must be P1D when ChargeType is Fee

Unknown value added to Resolution to avoid default values ending up as valid input

## References

* #195

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/Resolution.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/Resolution.cs
@@ -20,6 +20,7 @@ namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
     /// </summary>
     public enum Resolution
     {
+        Unknown = 0,
         PT15M,
         PT1H,
         P1D,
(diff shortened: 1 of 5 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/Resolution.cs
```diff
@@ -20,6 +20,7 @@ namespace GreenEnergyHub.Charges.Domain.ChangeOfCharges.Transaction
     /// </summary>
     public enum Resolution
     {
+        Unknown = 0,
```
bemwen, line 23: We can look past it right now, but note that this is actually a dangerous change going forward, as it moves the definition of PT15M from 0 to 1 in the data model.

bemwen (approved):

lasrinnil (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/ChangeOfCharges/Transaction/Resolution.cs
BjarkeMeier: How about considering using the names instead of the numbers?

PR #253, opened 2021-05-20
Change of charges integration test11

HenrikSommer:
The outcome of this PR will be a separate CI build intended integration test, that can be used for integration test after each function deploy. This is our preferred solution for now.

The PR also contains a less effective but probably functioning build of integration test for Command Receiver. We are aiming to remove this in a later PR, if the preferred solution is functional.

### DIFF
--- a/.github/workflows/integration-test-ci.yml
+++ b/.github/workflows/integration-test-ci.yml
@@ -1,0 +1,73 @@
+# Copyright 2020 Energinet DataHub A/S
+#
+# Licensed under the Apache License, Version 2.0 (the "License2");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+name: Integration Test CI
+
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch: {}
+
+env:
+  CSPROJ_FILE_PATH: 'source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/GreenEnergyHub.Charges.IntegrationTests.csproj'
+  SOLUTION_FILE_PATH: source/GreenEnergyHub.Charges/GreenEnergyHub.Charges.sln
+  DOTNET_VERSION: '3.1'
+
+jobs:
+  pre_job:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - name: Should skip?
+        id: skip_check
+        uses: fkirc/skip-duplicate-actions@v1.4.0
+        with:
+          github_token: ${{ github.token }}
+          paths: '[
+            "source/GreenEnergyHub.Charges/**",
+            ".github/workflows/integration-test-ci.yml",
+            ".github/actions/dotnet-build-and-test/**",
+            ".github/actions/azure-core-tools-install/**"
+          ]'
+
+  validate_integration_test:
+    needs: pre_job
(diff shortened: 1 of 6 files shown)


bemwen (approved):
LGTM, besides the danger of reading drop-it as something other than drop-integration-tests

HenrikSommer:
> LGTM, besides the danger of reading drop-it as something other than drop-integration-tests

Good point, it will be renamed in a later PR.

.github/workflows/integration-test-ci.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Should skip?
+        id: skip_check
+        uses: fkirc/skip-duplicate-actions@v1.4.0
+        with:
+          github_token: ${{ github.token }}
+          paths: '[
+            "source/GreenEnergyHub.Charges/**",
+            ".github/workflows/integration-test-ci.yml",
+            ".github/actions/dotnet-build-and-test/**",
+            ".github/actions/azure-core-tools-install/**"
```
BjarkeMeier, line 42: nit: I would still leave this one out (the azure-core-tools-install)

.github/workflows/integration-test-ci.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Setup .NET ${{ env.DOTNET_VERSION }} environment
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+
+      - name: Setup Azure Core Tools
+        uses: ./.github/actions/azure-core-tools-install
+
+      - name: Build and test project
+        uses: ./.github/actions/dotnet-build-and-test
```
BjarkeMeier, line 63: I don't think you should execute the tests here. I know there are none that will be executed but using this action seems to tell a misleading story to the reader.

BjarkeMeier (approved):
Approve, but with comments

.github/workflows/integration-test-ci.yml
```diff
@@ -0,0 +1,73 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch: {}
+
+env:
+  CSPROJ_FILE_PATH: 'source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/GreenEnergyHub.Charges.IntegrationTests.csproj'
+  SOLUTION_FILE_PATH: source/GreenEnergyHub.Charges/GreenEnergyHub.Charges.sln
```
Mech0z, line 24: Can these be changed to the same style as in .github/workflows/charge-command-receiver-cd.yml
          
```
          CSPROJ_FILE_PATH: ${{ env.INTEGRATION_TEST_CSPROJ_FILE_PATH }}
          SOLUTION_FILE_PATH: ${{ env.SOLUTION_FILE_PATH }}
```

Mech0z (approved):

.github/workflows/integration-test-ci.yml
BjarkeMeier: Aren't you mixing up the declaration and the usage?

.github/workflows/integration-test-ci.yml
HenrikSommer: So, you would rather just 
dotnet build ${{ csproj_file_path }} --configuration Release --output ${{ output_path }}
and not use the dot-net-and-test action?

.github/workflows/integration-test-ci.yml
HenrikSommer: Each CI and CD workflow configures its own ENV, like this:
env:
  CSPROJ_FILE_PATH: 'source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.ChargeCommandReceiver/GreenEnergyHub.Charges.ChargeCommandReceiver.csproj'
  SOLUTION_FILE_PATH: 'source/GreenEnergyHub.Charges/GreenEnergyHub.Charges.sln'

.github/workflows/integration-test-ci.yml
BjarkeMeier: Yes, because otherwise, I would read your code as if you want to execute the tests. I might even be at risk of trying to "fix" that the tests apparently don't run.

PR #276, opened 2021-05-25
Renamed ValidationRuleIdentifier's values.

lasrinnil:
## Description

During our latest (and first) team dev. session. We discussed our ValidationRuleIdentifier's values. Some were describing the name of the rule that failed, others were describing what was the error of the rule that failed. 

We have decided for now that for simplicity the ValidationRuleIdentifier's values now should describe the name of the rule that failed. 

## References

* #0000


Mech0z (approved):

HenrikSommer (approved):

BjarkeMeier (commented):
nit: There seem to be two different naming conventions for `ValidationRuleIdentifier` literals. With and without a `Validation`suffix.

PR #318, opened 2021-06-07
test: InputValidationRulesFactory

lasrinnil:
This PR. will add unittests to to the `InputValidationRulesFactory`. Making sure if any rules are removed from the implementation of either create / update / stop the corresponding test will fail.


bemwen (approved):

BjarkeMeier (approved):
LGTM

I notice that you have selected a different approach from the unit tests of the business validation rules factories. It's probably nit but I prefer to be able to see the expected values in the tests and not have to find them in a auxiliary private method. Also, when suitable I like tests working on single expected values rather than collections.

Another note: Neither of us actually test that the returned collection of actual rules ONLY contains the expected rules.

Feel free to merge :-)

BjarkeMeier:
Btw, thx for fixing this missing code coverage!

lasrinnil:
> LGTM
> 
> I notice that you have selected a different approach from the unit tests of the business validation rules factories. It's probably nit but I prefer to be able to see the expected values in the tests and not have to find them in a auxiliary private method. Also, when suitable I like tests working on single expected values rather than collections.
> 
> Another note: Neither of us actually test that the returned collection of actual rules ONLY contains the expected rules.
> 
> Feel free to merge :-)

I have merged the tests that ensure that each commands rules along with the mandatory rules were present. The new tests now ensures that the rules are present, and only them :)

PR #319, opened 2021-06-07
Add pipeline checking the charge solution using SonarCloud

bemwen:
## Description

The purpose of this PR is to integrate the SonarCloud controls into the charges domain.

The controls will be used, but the QualityGate in SonarCloud will be configured to allow everything in regardless of result until the team decides on what measurements should be the baseline.

* Adds pipeline that checks PRs using SonarCloud

## References

* See https://sonarcloud.io/organizations/energinet-datahub/projects for GreenEnergyHub projects that are now tracked through SonarCloud

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,0 +1,55 @@
+name: SonarCloud
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - source/coordinator/**
+      - source/messaging/**
+      - .github/workflows/sonarcloud.yml
+
+  pull_request:
+    types: [opened, synchronize, reopened]
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,55 @@
+name: SonarCloud
+on:
+  push:
+    branches:
+      - main
+    paths:
```
BjarkeMeier, line 6: Why not also `source/**` or at least `source/GreenEnergyHub.Charges/**` ?

BjarkeMeier (approved):
Nice :-)

.github/workflows/sonarcloud.yml
bemwen: Good point, changed

PR #324, opened 2021-06-08
test: ChargeFactory

lasrinnil:
To ensure that all properties can be mapped going from a `ChargeCommand` to `Charge`.
This PR will introduce a test for the `ChargeFactory`, so if `Charge` is expanded we wont forget map the new property in the `ChargeFactory`.

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Application/Factories/ChargeFactoryTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Application/Factories/ChargeFactoryTests.cs
@@ -1,0 +1,94 @@
+﻿// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


prtandrup (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Application/Factories/ChargeFactoryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    [UnitTest]
+    public class ChargeFactoryTests
+    {
+        [Fact]
+        public async Task CreateFromCommandAsync_Charge_HasNoNullsOrEmptyCollections()
+        {
+            // Arrange
+            var sut = new ChargeFactory();
+            var chargeCommand = new ChargeCommand("CorrelationId")
```
BjarkeMeier, line 36: Can't you have Moq instantiate a hydrated `ChargeCommand` using `InlineAutoMoqDataAttribute`?

lasrinnil: Ah I actually thought that meant that properties that were nullable would be null. But that is not the case, so I rewrote the test.

Mech0z (approved):

PR #326, opened 2021-06-08
Enable integration test execution from clone

HenrikSommer:
Being open source we strive to make it easy to clone, build and run all tests in the Charges domain. We therefore abort integration test execution if not configured. Furthermore we filter execution of tests integration instead of [Skip]'ing.


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/ChargeCommandReceiverConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.Azure.ServiceBus;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GreenEnergyHub.Charges.IntegrationTests.TestHelpers
+{
+    public class ChargeCommandReceiverConfiguration : ChargeCommandReceiver.Startup
+    {
+        private readonly ITopicClient? _topicClient;
+
+        public ChargeCommandReceiverConfiguration()
```
lasrinnil, line 32: Is the empty constructor needed?

lasrinnil (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargeLocalHostTests.cs
```diff
@@ -69,8 +70,17 @@ public ChangeOfChargeLocalHostTests(ITestOutputHelper testOutputHelper, [NotNull
             _testOutputHelper = testOutputHelper;
             _chargeDbQueries = new ChargeDbQueries(dbContextRegistrator.ServiceProvider);
 
-            var messageReceiverHost = FunctionHostConfigurationHelper.SetupHost(new MessageReceiver.Startup());
-            var chargeCommandReceiverHost = FunctionHostConfigurationHelper.SetupHost(new ChargeCommandReceiver.Startup());
+            var environmentVariables = Environment.GetEnvironmentVariables();
+            var environmentVariableReader = new EnvironmentVariableReader(environmentVariables);
+            _runLocalhostTests = environmentVariableReader.GetEnvironmentVariableOrFalse("RUN_LOCALHOST_TESTS");
```
BjarkeMeier, line 75: I find it harder than necessary to understand the concept about having to instantiate everything after this point when `_runLocalhostTests == false`. You said it was to align with upcoming squadron but as I understand @Sondergaard, then squadron will not be used for this type of test.

Can we refactor somehow to make the code and design more intuitive?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargeLocalHostTests.cs
```diff
@@ -82,25 +92,28 @@ public ChangeOfChargeLocalHostTests(ITestOutputHelper testOutputHelper, [NotNull
...
+            _commandAcceptedTopicName = environmentVariableReader.GetEnvironmentVariableOrEmptyString("COMMAND_ACCEPTED_TOPIC_NAME");
+            _commandRejectedTopicName = environmentVariableReader.GetEnvironmentVariableOrEmptyString("COMMAND_REJECTED_TOPIC_NAME");
+            _commandReceivedConnectionString = environmentVariableReader.GetEnvironmentVariableOrEmptyString("COMMAND_RECEIVED_LISTENER_CONNECTION_STRING");
+            _commandAcceptedConnectionString = environmentVariableReader.GetEnvironmentVariableOrEmptyString("COMMAND_ACCEPTED_LISTENER_CONNECTION_STRING");
+            _commandRejectedConnectionString = environmentVariableReader.GetEnvironmentVariableOrEmptyString("COMMAND_REJECTED_LISTENER_CONNECTION_STRING");
         }
 
-        [Theory(Skip = "Run at localhost to integration test local code using sandboxed Service Bus")]
+        [Theory]
+        [Trait("HostingEnvironment", "LocalHost")]
```
BjarkeMeier, line 107: What does this do? How about introducing constants with code comment. Then we can avoid misspelling and wrong casing. Also, we could add documentation for the trait.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargePipelineTests.cs
```diff
@@ -61,12 +65,15 @@ public ChangeOfChargePipelineTests([NotNull] DbContextRegistrator dbContextRegis
         }
 
         [Theory(Timeout = 60000)]
+        [Trait("HostingEnvironment", "Development")]
```
BjarkeMeier, line 68: See my comment about other trait about.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestFiles/InvalidCreateTariffCommand.json
```diff
@@ -16,7 +16,7 @@
     }
   },
   "ChargeOperation": {
-    "Id": "{{$isoTimestamp}}",
+    "Id": "MD{{$isoTimestamp}}",
```
BjarkeMeier, line 19: Just curious: Why the `MD`?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/EnvironmentVariableReader.cs
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections;
+
+namespace GreenEnergyHub.Charges.IntegrationTests.TestHelpers
+{
+    public class EnvironmentVariableReader
```
BjarkeMeier, line 19: Sorry to say, but I'm not really fond of this class for several reasons.

- The name suggests it's specific to environment variables, but it's not
- Instead of using `GetEnvironmentVariableOrEmptyString` I would just do something like `Environment.GetEnvironmentVariable("Foo") ?? string.Empty`
- And _if_ I wanted a utility method for doing the `?? string.Empty` I would put it in a much more generic utility class that wasn't tied to neither env variables nor dictionaries
- The `GetEnvironmentVariableOrFalse` and `GetEnvironmentVariableOrTrue` methods has a behavior that can't be guessed from neither the method names nor their code documentation

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/MessageReceiverConfiguration.cs
```diff
@@ -0,0 +1,69 @@
...
+                .AddMessaging()
+                .AddMessageExtractor<ChargeCommand>();
+
+            builder.Services
+                .AddSingleton(_ => _topicClient)
+                .AddScoped<IMessageDispatcher<ChargeCommandReceivedEvent>, MessageDispatcher<ChargeCommandReceivedEvent>>()
+                .AddScoped<Channel<ChargeCommandReceivedEvent>, TopicChannel>();
+        }
+
+        private class TopicChannel : Channel<ChargeCommandReceivedEvent>
```
BjarkeMeier, line 54: Why is this new class needed?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/ChargeCommandReceiverConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+using GreenEnergyHub.Charges.Domain.Events.Local;
+using GreenEnergyHub.Charges.Infrastructure.Messaging;
+using GreenEnergyHub.Charges.Infrastructure.Messaging.Registration;
+using Microsoft.Azure.Functions.Extensions.DependencyInjection;
+using Microsoft.Azure.ServiceBus;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GreenEnergyHub.Charges.IntegrationTests.TestHelpers
+{
+    public class ChargeCommandReceiverConfiguration : ChargeCommandReceiver.Startup
```
BjarkeMeier, line 28: What is the idea of deriving? And why can't we just reuse the existing Startup?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/ChargeCommandReceiverConfiguration.cs
HenrikSommer: For now it isn't needed. I have included it for preparation of integration tests using Squadron

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargeLocalHostTests.cs
HenrikSommer: Instantiating objects after this point is really only done so that we avoid changing them to be nullable.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargeLocalHostTests.cs
HenrikSommer: Traits enable filtering tests in our pipeline:
dotnet test ${{ inputs.test_dll }} --filter HostingEnvironment=Development

I will introduce constants.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/EnvironmentVariableReader.cs
HenrikSommer: No problem, I will move on with the `Environment.GetEnvironmentVariable("Foo") ?? string.Empty` style.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/MessageReceiverConfiguration.cs
HenrikSommer: When executing integration tests at localhost without any "local.setting.json" file defined, then the StartUp class will throw exceptions due to missing environment variables: "Function app is missing required environment variable"

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestHelpers/ChargeCommandReceiverConfiguration.cs
HenrikSommer: Same answer as above.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestFiles/InvalidCreateTariffCommand.json
HenrikSommer: Because OriginalTransactionReferenceMRid at the receivedEvent is prefixed with "MD"

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargeLocalHostTests.cs
BjarkeMeier: It's not a big deal, but yes I'd prefer that. I always expect that there is a reason for code to exist and for code to execute. When that's not true I get confused because I'm assuming something that I think should always hold true but apparently doesn't.

So I'd prefer the null solution or at least a comment in the code everywhere (only one location?) where code is added/executed just to avoid nulls.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/TestFiles/InvalidCreateTariffCommand.json
BjarkeMeier: I don't understand, but that's probably ok.

BjarkeMeier (approved):

Issue #362, opened 2021-06-23
Charge Links: (BRS-037) Link Receiver Returning 200 OK


Mech0z:
Synopsis
A HTTP trigger that receives an XML payload and returns 200 OK

Scope
At this point we do not need to understand the CIM/XML or process its content

ACs

- Valid Charge Link Created Postman script with CIM/XML links added
- Using postman a 200 OK is received

Tech Note
@prtandrup have knowledge on how to get a valid CIM XML

![image.png](https://images.zenhubusercontent.com/60bdbd325b600c62f9576fbb/b7e9a9de-b125-4fe0-838b-2d32d2db95dd)

prtandrup:
**Tested OK against T environment:**

**AC1**: Valid script available in Postman (Charges workspace)
**AC2**: ChargelinkHttpTrigger (Charge link receiver) returns '200 OK'

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/f6371c39-0a7e-48ec-8baf-f6d854d0c65b)

PR #377, opened 2021-06-29
Add Shell for Function receiving Charge Links

bemwen:
## Description

The purpose of this PR is to add the shell of the Azure Function will be receiving charge links.

The function itself is lightweight for now, just returning HTTP 200 to anything sent at it.

The project and pipelines utilize .NET 5.0 in preparation for charges moving to .NET 5.0 entirely later.

This PR adds:

* ChargeLinkReceiver project with the function
* Terraform for deploying the function
* Update to our deployment from local script to support the new function
* A change to the deployment from local script to use correct variable name
* CI pipeline for the ChargeLinkReceiver
* CD pipeline for the ChargeLinkReceiver

## References

* #362

### DIFF
--- a/.github/workflows/link-receiver-cd.yml
+++ b/.github/workflows/link-receiver-cd.yml
@@ -1,0 +1,122 @@
+# Copyright 2020 Energinet DataHub A/S
+#
+# Licensed under the Apache License, Version 2.0 (the "License2");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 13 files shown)


prtandrup (approved):
Looks good to me 👍 Think we should be renaming MessageReceiver, so it explicitly states what it receives, like the ChargeLinkReceiver does. Out of scope here though.

Mech0z (approved):

Issue #363, opened 2021-06-23
Metering Point: (BRS-004) MP Created Receiver


Mech0z:
Create a EventHub trigger capable of receiving a BRS-004 using Protobuf

**ACs**

Given a MeteringPointCreatedEvent is published in U environment by Charges domain
When consuming this event
Then we log the event content (where?)

Given a MeteringPointCreatedEvent is published in T environment Metering Point domain
When consuming this event
Then we log the event content (where?)

![image.png](https://images.zenhubusercontent.com/60bdd871726f411a853d94a9/139230bc-78d4-45d5-a76e-31e8a4c11cdf)

BjarkeMeier:
@Mech0z, should dashboard health monitoring of the new function have an AC?

Mech0z:
https://github.com/Energinet-DataHub/geh-charges/tree/msp/feat_358_ChargeLinkCommandReceiver

PR #382, opened 2021-07-01
Add infrastructure to support subscribing to metering point created events

BjarkeMeier:
## Description

Introduces:
- Concept for supporting infrastructure variations across environments
  Most importantly this is the first example of how we can isolate the domain from external dependencies like shared resources and other domains
- Infrastructure file organization has been revamped
- Added pipelines to set up and tear down an environment for each PR testing terraform configuration

After a talk with the software architect, it has been decided to go with a Service Bus queue rather than a topic with a subscription. That will be addressed in a later PR.

## References

* #363


build/infrastructure/.gitignore
```diff
@@ -23,4 +23,5 @@
 *.tfvars
 *.tfstate
 *.tfstate.backup
-*.backend.tf.exclude
+tfplan
+.working-folder
```
bemwen, line 27: Nice, well remembered

Mech0z (approved):

bemwen (approved):

PR #412, opened 2021-07-12
use shared GitHub actions

Renetnielsen:
## References

* #0000


Renetnielsen:
Change is to big compared to main, i will disband this and create new PR

PR #439, opened 2021-07-15
build: add sb topic auth rule listener

Mech0z:
## Description

We also need a listener to make the subscription useful

## References

* #0000

### DIFF
--- a/build/infrastructure/main/sb-charges.tf
+++ b/build/infrastructure/main/sb-charges.tf
@@ -128,6 +128,16 @@ resource "azurerm_servicebus_subscription" "sbs-link-command-received" {
 
 module "sbtar_command_rejected_listener" {
   source                    = "git::https://github.com/Energinet-DataHub/geh-terraform-modules.git//service-bus-topic-auth-rule?ref=1.3.0"
+  name                      = "sbtar-link-command-received-listener"
+  namespace_name            = module.sbn_charges.name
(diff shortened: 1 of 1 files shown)


build/infrastructure/main/sb-charges.tf
```diff
@@ -126,6 +126,16 @@ resource "azurerm_servicebus_subscription" "sbs-link-command-received" {
   max_delivery_count  = 1
 }
 
+module "sbtar_command_rejected_listener" {
```
prtandrup, line 129:
```suggestion
module "sbtar_link_command_received_listener" {
```

HenrikSommer, line 129: I assume this should have been named "sbtar_command_received_listener"?

lasrinnil, line 129: 2 modules with the same name is this correct? (I don't really know)

lasrinnil (approved):

build/infrastructure/main/sb-charges.tf
Mech0z: Fixed in 8b35a70e0aa3aae464d4afb6b2eec1bbab21d9ac

Mech0z: Fixed in 8b35a70e0aa3aae464d4afb6b2eec1bbab21d9ac

prtandrup (approved):

PR #468, opened 2021-07-27
refac: Remove use of Jetbrains.Annotation

bemwen:
## Description

The purpose of this PR is to remove our uses of the JetBrains.Annotation DLL.

We are currently using it for [UsedImplicitly] and often misusing it for [NotNull].

It will be easier to spot if we dont use the DLL at all.


BjarkeMeier (approved):
How do we now obtain a green R#?

Mech0z (approved):

bemwen:
> How do we now obtain a green R#?

If that is a goal for us, we should probably make an issue with analysing it. There are other things that it highlights at the moment that we should also address in that case, like our InlineAutoMoqData based tests which also gets a yellow remark at the moment.

BjarkeMeier:
> > How do we now obtain a green R#?
> 
> If that is a goal for us, we should probably make an issue with analysing it. There are other things that it highlights at the moment that we should also address in that case, like our InlineAutoMoqData based tests which also gets a yellow remark at the moment.

From my personal experience, a "green R#" requirement is very simple to understand and follow and often leads to a more homogenous code style across developers. Especially when combined with using the "clean up code" feature.

Issue #143, opened 2021-04-27
Notify the team if MessageReceiver isn't responding correctly


PerLysemose:
In order to ensure working software we take a step in direction of surveilling health state of our Azure functions.

**Tech notes**

Energinet has shut down webhooks of Teams channels due to security concerns. So it is assumed that the Microsoft Teams email is being set up in GitHub repo secrets settings and passed on to Terraform through the GitHub infrastructure pipeline.
Teams channel email can be found in the teams channel you want to target.

**Acceptance criteria**

1. In case one of the existing availability tests in the charges domain fail a notification is shown in team Volts channel in Microsoft Teams

BjarkeMeier:
Original text:

_It is fairly easy to add tooling to continously monitor the message receiver to verify that it is responding to HTTP requests. E.g. using availability tests deployed using Terraform. These can send mails or other kinds of notification. Wouldn't it be nice to be notified within 5ish minutes when the endpoint is down - e.g. due to breaking PRs or other reasons?_

bemwen:
Note: Availibility test of the time series domain has been postponed and moved to #471

The AC has been updated to reflect this by added charges as the only target.

bemwen:
To the tester.

You can test an alert by simply stopping an Azure Function and waiting for the alert to raise.

prtandrup:
**Tested :white_check_mark: in T environment.** 

Turned off `message-receiver-charges`:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/c4b9d829-c91c-44b2-9611-dafa3892b564)

Notifications sent to, and shown in Teams:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/66c1af78-288e-4c39-ae97-9bc56accfca8)

PR #469, opened 2021-07-27
build: Add availability terraform modules

bemwen:
## Description

The purpose of this PR is to change and add new modules to terraform that we can use to add alerts to our availability tests later.

Use of the modules will not be added until a later PR.

This PR:
* Adds a module for adding a monitoring alert group
* Adds a module for adding alerts of availability tests
* Adds extra output to the ping-test-module needed by the new modules
* Removes our own storage account module which is no longer used

## References

* #143


Mech0z (approved):

BjarkeMeier (commented):
I know we don't tend to write documentation in our infra code, but isn't that just a bad habit?

If we put this stuff in this repo then we'll end up with either having to move and refactor later, or duplicate in other repos later. Also, other teams are not likely to get to know and benefit from our work unless we put it in https://github.com/Energinet-DataHub/geh-terraform-modules. And yes, it's probably me who started some of the work here, and yes MightyDucks are on vacation...

BjarkeMeier (approved):

PR #472, opened 2021-07-28
Add alerts for availability tests

bemwen:
## Description

The purpose of this PR is to add alerts for availability tests to our Azure Functions though Terraform.

This PR:
* Adds an alert for each of our 8 current functions in the charge domain
* Adds the alert group which receives the alerts

## References

* #143

### DIFF
--- a/build/infrastructure/main/azfun-charge-confirmation-sender.tf
+++ b/build/infrastructure/main/azfun-charge-confirmation-sender.tf
@@ -81,3 +81,18 @@ module "ping_webtest_charge_confirmation_sender" {
   url                             = "https://${module.azfun_charge_confirmation_sender.default_hostname}/api/HealthStatus"
   dependencies                    = [module.azfun_charge_confirmation_sender.dependent_on]
 }
+
+module "mma_ping_webtest_charge_confirmation_sender" {
+  source                   = "../modules/availability-alert"
+  name                     = "mma-charge-confirmation-sender-${var.project}-${var.organisation}-${var.environment}"
+  resource_group_name      = data.azurerm_resource_group.main.name
+  application_insight_id   = module.appi.id
+  ping_test_name           = module.ping_webtest_charge_confirmation_sender.name
+  action_group_id          = module.mag_availabilitity_group.id
+  tags                     = data.azurerm_resource_group.main.tags
+  dependencies             = [
+    module.appi.dependent_on,
+	module.ping_webtest_charge_confirmation_sender.dependent_on,
+	module.mag_availabilitity_group.dependent_on
+  ]
+}
--- a/build/infrastructure/main/azfun-charge-confirmation-sender.tf
+++ b/build/infrastructure/main/azfun-charge-confirmation-sender.tf
@@ -92,7 +92,7 @@ module "mma_ping_webtest_charge_confirmation_sender" {
   tags                     = data.azurerm_resource_group.main.tags
   dependencies             = [
     module.appi.dependent_on,
-	module.ping_webtest_charge_confirmation_sender.dependent_on,
-	module.mag_availabilitity_group.dependent_on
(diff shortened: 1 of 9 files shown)


Mech0z (approved):

build/infrastructure/main/azfun-charge-confirmation-sender.tf
```diff
@@ -81,3 +81,18 @@ module "ping_webtest_charge_confirmation_sender" {
...
+  name                     = "mma-charge-confirmation-sender-${var.project}-${var.organisation}-${var.environment}"
+  resource_group_name      = data.azurerm_resource_group.main.name
+  application_insight_id   = module.appi.id
+  ping_test_name           = module.ping_webtest_charge_confirmation_sender.name
+  action_group_id          = module.mag_availabilitity_group.id
+  tags                     = data.azurerm_resource_group.main.tags
+  dependencies             = [
+    module.appi.dependent_on,
+    module.ping_webtest_charge_confirmation_sender.dependent_on,
+    module.mag_availabilitity_group.dependent_on
```
BjarkeMeier, line 96: I know it seems to be practice in GEH. But I think that the `.dependent_on` isn't needed and is a habit that should be dropped. I mentioned it briefly to @Renetnielsen a while ago and I _think_ he tended to agree.

build/infrastructure/main/mag-availability.tf
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module "mag_availabilitity_group" {
+  source              = "../modules/monitor-action-group-email" # Repo geh-terraform-modules doesn't have a monitor-action-group at the time of writting this
+  name                = "mag-availability-${var.project}-${var.organisation}-${var.environment}"
+  resource_group_name = data.azurerm_resource_group.main.name
+  short_name          = "a-${var.project}-${var.environment}"
```
BjarkeMeier, line 19:
```suggestion
  short_name          = "a-${var.project}-${var.organisation}-${var.environment}"
```

BjarkeMeier (approved):

build/infrastructure/main/azfun-charge-confirmation-sender.tf
```diff
@@ -81,3 +81,18 @@ module "ping_webtest_charge_confirmation_sender" {
   url                             = "https://${module.azfun_charge_confirmation_sender.default_hostname}/api/HealthStatus"
   dependencies                    = [module.azfun_charge_confirmation_sender.dependent_on]
 }
+
+module "mma_ping_webtest_charge_confirmation_sender" {
```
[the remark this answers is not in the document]
bemwen: monitor metric alert. We usually prefix with the abbreviated name of the resource the module is using.

build/infrastructure/main/mag-availability.tf
bemwen: That would make the name to long, only 12 characters are allowed

build/infrastructure/main/azfun-charge-confirmation-sender.tf
bemwen: Nod, we need to make sure that is the case though, and I would rather not make it a part of this PR. In my experience, a trap can also be that it works during setup and fails during teardown, where dependencies can again make a difference. We will need to make sure this is not one of those cases.

build/infrastructure/main/mag-availability.tf
BjarkeMeier: Then perhaps a little doc to explain this deviation :-)

PR #486, opened 2021-08-09
build: PR env setup to only run on infrastructure or code changes in charges repo

prtandrup:
## Description

This PR changes the workflow 'PR environment - setup', so it checks whether infrastructure, workflow or charges code base is changed. If just one of them gets changed, the workflow runs.

This way it won't run upon e.g. documentation changes. 

## References

* #0000


bemwen (approved):

.github/workflows/pr-env-setup.yml
```diff
@@ -35,8 +35,43 @@ env:
...
+          paths: '[
+            "source/**",
+            ".github/workflows/pr-env-setup.yml"
+          ]'
+
+  # Infrastructure deployment is only needed if there have been changes to infrastructure, the charges code, or this workflow.
   infra_deploy:
     name: Deploy infrastructure
+    needs: [infra_pre_job, charges_pre_job]
+    if: ${{ needs.infra_pre_job.outputs.should_skip != 'true' || needs.charges_pre_job.outputs.should_skip != 'true' }}
```
BjarkeMeier, line 74: The two jobs define a "should_skip" prop, but they don't who will use them and for what. Instead the props should be declarative like e.g. "has_changes". That would

* Make them reusable for other purposes
* Ensure that the maintainer doesn't have to know the context/use of the pre jobs in order to understand the props and their naming
* This if-clause would be much easier to read and save the two negations

prtandrup: I agree that should_skip is not optimal, but it looks like the way [skip-duplicate-actions](https://github.com/fkirc/skip-duplicate-actions#outputs) has implemented it. So I prefer we stick with that. Agree?

BjarkeMeier: The `skip-duplicate-actions` action does not need to and should not leak out of the pre-job.

BjarkeMeier: I'm changing it for the time series domain [here](https://github.com/Energinet-DataHub/geh-timeseries/blob/Use_protobuf_for_received_time_series/.github/workflows/pr-env-setup.yml).

prtandrup: Yeah, I saw that :) Will be reflected here once it's in time series main.

prtandrup: I've aligned with time series domain, so it outputs "has_changes" now.

BjarkeMeier (approved):

Issue #373, opened 2021-06-28
Charges: Use protobuf for events sent internally


bemwen:
**Synopsis**

As per the ADR decision, we need to move our communication from JSON-based to protobuf-based internally in the Charge domain.

**Out of scope**
Messages sent to our own internal post office which we will need to replace is out of scope of this task.

**Acceptance criteria**

1. ChargeCommands sent between MessageReceiver and ChargeCommandReceiver must use Protobuf
2. ChargeCommandAccepted events sent between the ChargeCommandReceiver and the ChargeConfirmationSender must use Protobuf
3. ChargeCommandRejected events sent between the ChargeCommandReceiver and the ChargeRejectionSender must use protobuf

**Testing**

1. Send a valid Charge HTTP Trigger "Create Tariff message" using postman and validate that a Protobuf message is received in topic sbs-post-office
2. Send a invalid by changing StartDateTime to a date less than 30 days from now and see that a Protobuf message is received containing "StartDateValidation" meaning it failed validation in topic sbs-post-office

prtandrup:
Tested :white_check_mark: in T environment

````
Test procedure
- Disable the following functions: ChargeCommandReceiver and ChargeConfirmationSender
- Submit valid charge message via Postman to the MessageReceiver
- Inspect (peek) the service bus subscription for command-received (and verify content is protobuf, or at least not plain JSON)
- Enable the ChargeCommandReceiver
- Inspect (peek) the service bus subscription for command-accepted (protobuf, not plain JSON)
- Enable the ChargeConfirmationSender
- Inspect (peek) the service bus subscription for post-office (protobuf, not plain JSON)
- Using the hackathon project AzureProtobufHelper to read from the post office subscription, and verify the readable content in the console application

Test procedure can be repeated for the invalid scenario replacing ChargeConfirmationSender with ChargeRejectionSender and submitting an invalid charge message. 
````
AzureProtobufHelper project [link](https://github.com/Energinet-DataHub/hackathon-projects/tree/master/AzureProtobufHelper)

**Test: valid charge message**:
Submitted valid charge message, got correlation id = e6ea2c4a-0c56-4189-aaef-c7082ce4f6f5
Verified message in `command-received` is proto, not plain JSON: 
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/5c00a156-3be1-473e-a68b-2ccffcff5968)

[...]

PR #491, opened 2021-08-13
Add pipelines for create link command receiver

bemwen:
## Description

The purpose of this PI is to add CI/CD pipelines for the new create link command receiver

* Adds CI pipeline
* Adds CD pipeline

## References

* #373


HenrikSommer (approved):

.github/workflows/create-link-command-receiver-cd.yml
```diff
@@ -0,0 +1,122 @@
...
+          {
+            long: rg-DataHub-Charges-B,
+            short: b,
+            name: Preprod
+          },
+          {
+            long: rg-DataHub-Charges-P,
+            short: p,
+            name: Production
+          }
```
BjarkeMeier, line 92: Should we keep adding pipeline matrixes after @Renetnielsen has discouraged it and removed it from infra pipelines?

BjarkeMeier (approved):

.github/workflows/create-link-command-receiver-cd.yml
bemwen: Probably not. I think we should align best practice with him based on the new strategy. But as I running low on time, I would prefer to postpone this to another PR.

PR #550, opened 2021-09-03
Update issue templates

PerLysemose:


.github/ISSUE_TEMPLATE/Feature_Request.md
```diff
@@ -1,18 +1,31 @@
...
+title: ''
+labels: epic, feature
+assignees: ''
+
 ---
 
-### Is your feature request related to a problem? Please describe.
-<!--- A clear and concise description of what the problem is. Ex. I have an issue when [...] --->
+## Problem Description
+New Product: “The current state of the [domain] has focussed primarily on [customer segments, pain points, etc]. What existing products/services fail to address is [this gap] Our product/service will address this gap by [vision/strategy] Our initial focus
```
lasrinnil, line 11: Is it the intent that the description of the header is visible? I would prefer if it was only showed upon the creation of a PR and not a part of the actual PR description after it is created. 

If the descrption only should be visible upon creation, you can add following symbols:

PerLysemose: Yes, that is the intent. Format of template is decided by ART team so it will be the same for all repos

HenrikSommer (approved):

Mech0z (approved):

Issue #323, opened 2021-06-08
Extend existing integration tests with a check that charge prices were added to the database


prtandrup:
Description:
To be sure that both charge master data and prices are stored, the existing integration tests must be extended with an assert that verifies that charge prices were stored.

Acceptance criteria:
AC#1: Integration tests that integrate with Charges database are extended with an assert that verifies that Charge prices were stored
AC#2: Repository test is extended to assert charge prices were stored

HenrikSommer:
AC#2 is added following comments to pull request

lasrinnil:
###Test

I have verified that upon creation and retrieval of a charge, that points are present:
![image.png](https://images.zenhubusercontent.com/60bdfcb3effb2082025638c1/04be3ea5-6c3a-4c35-9964-4abb269482a1)

I have verified that the integration test which is run upon deploying the charge receiver now checks for points:
![image.png](https://images.zenhubusercontent.com/60bdfcb3effb2082025638c1/48de5c0a-ff03-44c6-a07b-5c0e2488f684)

PR #560, opened 2021-09-07
Check that charge prices were added to db in pipeline test

HenrikSommer:
With this PR we ensure that charge prices are added to the database when a charge is saved.

## References
#323


BjarkeMeier (commented):
Why is this part of the integration test rather than a -- much simpler to maintain -- repository test?

Mech0z (approved):

HenrikSommer:
> Why is this part of the integration test rather than a -- much simpler to maintain -- repository test?

Good question! - we defined story #323 three months ago, but maybe it is just not relevant anymore?

Mech0z (approved):

BjarkeMeier:
> > Why is this part of the integration test rather than a -- much simpler to maintain -- repository test?
> 
> Good question! - we defined story #323 three months ago, but maybe it is just not relevant anymore?

I would prefer a database unit test (and an update of the wording in the story)...

lasrinnil (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Application/ChangeOfCharges/ChangeOfChargePipelineTests.cs
```diff
@@ -95,18 +96,19 @@ public ChangeOfChargePipelineTests([NotNull] DbContextRegistrator dbContextRegis
...
             // assert
-            Assert.True(changeOfChargesMessageResult.IsSucceeded);
-            Assert.Equal(chargeCommand.ChargeOperation.Id, receivedEvent.OriginalTransactionReference);
-            Assert.NotNull(receivedEvent);
-            Assert.True(chargeExists);
+            changeOfChargesMessageResult.IsSucceeded.Should().BeTrue();
+            chargeCommand.ChargeOperation.Id.Should().Be(receivedEvent.OriginalTransactionReference);
+            receivedEvent.Should().NotBeNull();
+            charge.Should().NotBeNull();
+            charge.Points.Should().NotBeNullOrEmpty();
```
prtandrup, line 111: How about making this an explicit assert, checking that the price was in fact saved as provided in the test file?

Btw, I do see some values in the test file used, which seems postman related, e.g. 
"Time": "{{$isoTimestampPlusOneMonth}}" - Should we edit those values? If you'd like, I might have time for that today.

prtandrup:
> > > Why is this part of the integration test rather than a -- much simpler to maintain -- repository test?
> > 
> > 
> > Good question! - we defined story #323 three months ago, but maybe it is just not relevant anymore?
> 
> I would prefer a database unit test (and an update of the wording in the story)...

👍 for repo unit test
I'll like to keep this pipeline test in place. And instead raise the question whether this is an integration test or actually the beginning of how we can test flows throughout our domain. I will put a ticket up for our next dev session.

prtandrup (approved):

PR #561, opened 2021-09-07
Finalize renaming of MessageReceiver to ChargeReceiver

HenrikSommer:


BjarkeMeier (commented):
How about "ChargeMessageReceiver" to be explicitly distinguishable from e.g. "ChargeCommandReceiver" in order to help avoid confusion or mistakes?

How about adding "EntryPoints" in the namespace of all the hosts like in [metering point domain](https://github.com/Energinet-DataHub/geh-metering-point/tree/main/source)? Might be suitable for another PR.

HenrikSommer:
> How about "ChargeMessageReceiver" to be explicitly distinguishable from e.g. "ChargeCommandReceiver" in order to help avoid confusion or mistakes?
> 
> How about adding "EntryPoints" in the namespace of all the hosts like in [metering point domain](https://github.com/Energinet-DataHub/geh-metering-point/tree/main/source)? Might be suitable for another PR.

MessageReceiver was already renamed to ChargeReceiver, so this PR only finalizes this renaming. "ChargeMessageReceiver" might be a better name, but I think it will require a new team decision.

Mech0z (approved):

BjarkeMeier (approved):
I'm not in favour of the chosen name. But as this PR simply finishes something decided and partly implemented earlier I don't see any reason to not approve this PR.

Issue #555, opened 2021-09-07
Create new context for assert in database tests


Mech0z:
To ensure we are not just reading from memory in the old context, its been decided to create a new database context for the assert part of database tests

ACs
1. The context used to arrange and act is not the same as the one used to assert that data is valid

HenrikSommer:
[Pinal Dave - SQL-server login failures](https://logicalread.com/errors-sql-server-login-failures-pd01/#.YTfYdvdxeF4) describes a lot of SQL server login errors.
State 7 is the one in question here: 

> This would appear if _login is disabled_ AND _incorrect password is provided_.

BjarkeMeier:
Test of story:
- All tests using `SquadronContextFactory` have been manually confirmed to use different context for read and write
- I consider the integration tests that uses the `ChargeDatabaseContext` to be out of scope for this story
- All unit tests passes

Conclusion: Story is tested to be correct

PR #581, opened 2021-09-14
refac: use new context for reading database values

Mech0z:
## Description

To ensure we are not just reading from memory in the old context, its been decided to create a new database context for the assert part of database tests

ACs
1. The context used to arrange and act is not the same as the one used to assert that data is valid

## References

* #555

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure/Repositories/ChargeRepositoryTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure/Repositories/ChargeRepositoryTests.cs
@@ -15,6 +15,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Threading.Tasks;
 using GreenEnergyHub.Charges.Domain.Charges;
 using GreenEnergyHub.Charges.Domain.MarketDocument;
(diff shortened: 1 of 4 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure/Repositories/ChargeRepositoryTests.cs
```diff
@@ -50,62 +51,76 @@ public ChargeRepositoryTest(SqlServerResource<SqlServerOptions> resource)
...
+            await using var chargesDatabaseReadContext = await SquadronContextFactory
+                .GetDatabaseContextAsync(_resource)
+                .ConfigureAwait(false);
+            var expected = await chargesDatabaseReadContext.Charges.SingleAsync(x =>
+                x.ChargeId == charge.Id &&
+                x.MarketParticipant.MarketParticipantId == charge.Owner &&
+                x.ChargeType == (int)charge.Type)
+                .ConfigureAwait(false);
+
             Assert.NotNull(expected);
```
BjarkeMeier, line 74:
```suggestion
            var actual = await chargesDatabaseReadContext.Charges.SingleAsync(x =>
                x.ChargeId == charge.Id &&
                x.MarketParticipant.MarketParticipantId == charge.Owner &&
                x.ChargeType == (int)charge.Type)
                .ConfigureAwait(false);

            Assert.NotNull(actual);
```

BjarkeMeier: This "expected" vs "actual" naming fix is needed in multiple places.

BjarkeMeier (approved):
LGTM. It would be nice it you'll fix the expected/actual naming as you're touching the code - though I know they were not introduced in this PR.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure/Repositories/ChargeRepositoryTests.cs
Mech0z: fixed

HenrikSommer (approved):

BjarkeMeier:
CodeCov report doesn't look good. Why are coverage of all those files massively reduced?

PR #613, opened 2021-09-16
Fix integration test from clone

HenrikSommer:
With this pull request we reverse logic determining whether to run pipeline tests or not during development.
So that we can clone code and run all tests without any failing due to missing local.settings.json

This entails the need to provide an environment variable in our pipeline.


BjarkeMeier (commented):
What is now the recommended way of running these tests during development?

Mech0z (approved):

bemwen (approved):

HenrikSommer:
> What is now the recommended way of running these tests during development?

With today's commits to this pull request, a new attribute inherited from XUnit Theory will ensure that tests are skipped if no `integrationtest.local.settings.json` file is present, thus it is now possible to run all tests in the Charges domain right from Git clone.

As prior to this pull request, you can set up your own `integrationtest.local.settings.json` and run all integration tests using Azure sandbox resources.

Issue #516, opened 2021-08-20
EF configure datacontracts through FluentConfiguration


lasrinnil:
**Synopsis**
To better align with ART decisions we are going to remove our `Entity `models and configure our database contracts through `FluentConfiguration`.

Edit: Only Metering Point is to be ported as Charge will requires a bit refactoring which will happen at a later point https://github.com/Energinet-DataHub/geh-charges/issues/628

**Tech notes**
It is possible to configure EFCore to handle the Instant datatype see this GitHub repo for further instructions:
https://github.com/StevenRasmussen/EFCore.SqlServer.NodaTime

An example from Team Batman:
https://github.com/Energinet-DataHub/geh-metering-point/blob/main/source/Energinet.DataHub.MeteringPoints.Infrastructure/DataAccess/MeteringPoints/MeteringPointEntityConfiguration.cs

**Acceptance criteria**
~~1) No models are used only to configure DBSet.~~
2) All Unit/integration tests pass.
3) Metering Point is ported

BjarkeMeier:
@lasrinnil, where did the discussion about completely getting rid of the data model end?

BjarkeMeier:
Test of story:
- AC1: I accept that it has been deferred to #628
- AC2: Ensured by PR gate
- AC3: Verified by inspection

Conclusion: Story is tested OK.

PR #618, opened 2021-09-20
refac: change MP to use FluentConfiguration

Mech0z:
This changes Metering Point repository to use FluentConfiguration instead of entity mapping

## Description

To better align with ART decisions we are going to remove our `Entity `models and configure our database contracts through `FluentConfiguration`.

This is part of that, but not all as the Charges domain object is not straightforward to port to FluentConfiguration as the database and domain model do not match 1:1

## References

* #516


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/MeteringPoints/MeteringPoint.cs
```diff
@@ -34,24 +34,6 @@ public class MeteringPoint
             SettlementMethod = settlementMethod;
         }
 
-        public MeteringPoint(
```
BjarkeMeier, line 37: Isn't it more correct to remove the other ctor and remove the public setter of the `RowId` prop?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.ChargeCommandReceiver/Startup.cs
```diff
@@ -89,7 +90,7 @@ protected virtual void ConfigurePersistence([NotNull] IFunctionsHostBuilder buil
                                        "CHARGE_DB_CONNECTION_STRING",
                                        "does not exist in configuration settings");
             builder.Services.AddDbContext<ChargesDatabaseContext>(
-                options => options.UseSqlServer(connectionString));
+                options => options.UseSqlServer(connectionString, options => options.UseNodaTime()));
```
BjarkeMeier, line 93: Looks like you found a way better solution than what I did with the charge links. It would be nice to apply this everywhere and remove my extension methods for converting Instants.

BjarkeMeier (commented):
Looks good. Left a few comments.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/MeteringPoints/MeteringPoint.cs
Mech0z: Changed/readded to have the RowId? in the constructor 9ec8d167960ff4fb89ebb9c1fb547e8eacf11db7

BjarkeMeier: Doesn't look like the right commit(?) link.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.ChargeCommandReceiver/Startup.cs
Mech0z: refactored in f78c29fd2fde5ac0e316deeb65b3eb1b59f25916

BjarkeMeier (approved):
Great work @Mech0z

lasrinnil (approved):

Issue #334, opened 2021-06-09
Charge links: (BRS-037) Add/Store charge links upon a received charge link request


prtandrup:

BjarkeMeier:
Tasks
- [X] Add SQL model
- [X] Add create logic (no unit tests)
- [ ] Add create unit tests
- [ ] Add actual persistance
- [ ] Create story to use domain generated GUID ID values for entities everywhere
- [ ] Update ChargeLink data model entity diagram

BjarkeMeier:
Clean code tasks
- [ ] enddatetime ikke nullable
- [ ] Squadron problem med db unit tests
- [ ] MeteringPointRepository - dubleret kode - skal fokuseres på aggregate og ingen dublering
- [ ] Brug GUID ID alle steder
- [ ] Mangler vi nogle uniqueue indexes til at undgå ulovlige charge links? Charges.MeteringPointId, ..
- [ ] Fix connection state for dbup metetering point test data
- [ ] Organiser efter aggregates?
- [ ] Ingen integrationstests => nemt at breake main pga. DI (selvom al kode ellers er korrekt og med 100% coverage)

BjarkeMeier:
Removing this clean code task as it appears to probably not be worth the effort according to talk with @dstsyst about using integrationtests more intensively and actually spinning up function while executing XUnit unit tests (both localhost and in pipeline). Besides the assumed solution would be based on similar test in market role domain which happens to only test the actual configuration of the DI container but not whether all necessary dependencies has actually been configured (e.g. those from the function endpoints).

- [ ] Ingen integrationstests => nemt at breake main pga. DI (selvom al kode ellers er korrekt og med 100% coverage) (@BjarkeMeier)

prtandrup:
Testing in progress
Env: T
Commit:  [dcae461](https://github.com/Energinet-DataHub/geh-charges/commit/dcae461da0d6b3efc1ceb5c0b653ba31929e8b75)
<hr>

**AC1: Charge Links data model implemented and deployed:**

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/241f33d6-dd01-40fb-bc56-218ea567e451)

<hr>

**AC2: Using MP ID (571313180000000005) to create charge links:**
- Create tariff link (tariff: 334-tariff): 
CorrelationID: 9bcc1dbe-0929-47c1-9754-f283a81f1115
No EndDate provided in CIM XML-
[...]

PR #619, opened 2021-09-20
fix: Uncommented chargelink test

bemwen:
## Description

The purpose of this PR is work on the StoreAsync_StoresChargeLink test which is currently uncommented due to an unknown issue.

The problem is related to how quickly we use the SQL server within docker through Squadron, which Squadron has partly addressed in their upgrade to 0.12.1.

It is however not enough in our use scenario, so we increade the wait even.

We should consider whether it would be possible to make a better solution where we determine when the SQL server is ready to take on test, but that requires more knowledge of Squadron than we currently have.

This is a quick fix that work around the problem until we get an idea for a solution.

## References

* #334


HenrikSommer (approved):

BjarkeMeier (approved):
Great job.

prtandrup (approved):

PR #638, opened 2021-09-22
Clean up folder and file structure

BjarkeMeier:
## Description

Moved a lot of stuff back and forth between application and domain layers.
Test projects haven't been updated accordingly. I will wait for the final
structure to be settled and implemented, and team acceptance in order to avoid wasting work.

Domain project has been organized by domain aggregates.

For reviewers, it might be helpful to know that no implementation has been changed. The following should summarize what kind of changes have been made:

- Move files
- Create and delete folders
- Rename a few mappers to factory
- Add a reference from domain project to core project
- Adjust namespaces in domain and application projects

It might be easier to review by checking out and navigating around in domain and application projects:

image

image

## References

* #0000


bemwen (approved):
Great work!

This is a great first step in better structuring our code.

The validation rules being in domain instead of application, and whether events belong on the same level as aggregate roots are something I think we need to talk about sometime down the road, but this clearly moves us in the right direction.

Mech0z (approved):

PR #639, opened 2021-09-23
Improve misleading code coverage metrics

BjarkeMeier:
## Description

Our 94% code coverage is misleading as our functions endpoints are not included. This PR makes sure that they are included by adding a test that targets symbols from all the function hosts.

As an added benefit the test project - and thus also multiple pipelines - has been updated from .NET 3.1 to .NET 5.0.

## References

* #0000


prtandrup:
> ## Description
> Our 94% code coverage is misleading as our functions endpoints are not included. This PR makes sure that they are included by adding a test that targets symbols from all the function hosts.
> 
> As an added benefit the test project - and thus also multiple pipelines - has been updated from .NET 3.1 to .NET 5.0.
> 
> ## References
> * #0

It's not that I don't like the idea, but I'm somewhat confident that a decision was made earlier to not include the functions. Perhaps @bemwen do you remember

BjarkeMeier:
> It's not that I don't like the idea, but I'm somewhat confident that a decision was made earlier to exclude the functions. Perhaps @bemwen do you remember

If that's the case then it should be explicitly configured. Because as things are now they (soon: it) will automatically be included when someone for any reason adds a unit test testing something in the functions. Which should definitely be allowed.

bemwen (approved):

Mech0z (approved):

PR #701, opened 2021-10-12
Fix functionhost-ci.yml skip guard

BjarkeMeier:
## Description

PLEASE NOTE: Changed state to draft as it doesn't work as expected. Feel free to help :-) I've created an issue with request for help [here](https://github.com/dorny/paths-filter/issues/109).

The skip guard of build and test of functionhost CI workflow didn't work correctly resulting in irrelevant runs.
This PR should fix the problem.

If working properly we probably want to use the approach in all our pipelines with skip guards.

### DIFF
--- a/.github/workflows/functionhost-ci.yml
+++ b/.github/workflows/functionhost-ci.yml
@@ -46,23 +46,18 @@ jobs:
   pre_job:
     runs-on: windows-latest
     outputs:
-      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+      has_changes: ${{ steps.changes.outputs.src == 'true' }}
     steps:
-      - name: Should skip?
-        id: skip_check
-        uses: fkirc/skip-duplicate-actions@v1.4.0
+      - uses: dorny/paths-filter@v2
+        id: changes
         with:
-          github_token: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


BjarkeMeier:
The tested action doesn't work as expected and [my issue in their repo](https://github.com/dorny/paths-filter/issues/109#issuecomment-953311719) haven't been anwered within the last 20 days. Apparently someone left the following comment but deleted it again: "Agreed, I received an empty array instead of false too. Changing to checking for existence of the variable worked, but unfortunately my code will break if this action is updated to return false".

Conclusion: I don't consider the action to be stable enough or have a sufficiently active community to base our pipelines on it.

airtonix:
@BjarkeMeier you should revisit this or perhaps look at using NX.dev. It's affected functionality is amazing.

someone has created a github action that is like dorny/paths-filter, but instead it uses your nx project graph to answer the question, so there's no configuration required.

- https://github.com/marketplace/actions/nx-affected-list
- https://github.com/gagle/nx-check-changes
- there's another that I've looked at using but cant remember or find it.


You can also teach NXdev how to build a dep graph of any language. Here's an example done with golang: 

- https://github.com/nrwl/nx-go-project-graph-plugin

Issue #649, opened 2021-09-27
Provide charge links created bundles via post office (part 2/3): Create CIM XML bundle


BjarkeMeier:
** Synopsis **

Create bundle for the grid operator in order to support peek of charge link created messages in Green Energy Hub (the post office).

** Tech Notes **

See #642 for details.

** Acceptance-Criteria **

The bundle code will be used in part 3 when actually replying to the actors peek request.

AC 0 (technical requirement due to splitting of story):
Given the charges domain
When notifying the post office about created charge links
Then sufficient information is stored in order to be able to later create the bundled

AC 1: Code has been added that creates a CIM XML bundle from a set of ids (which in part 3/3 will be provided by the post office)
AC 2: The code is covered by automated test(s)

PR #732, opened 2021-10-21
feat: Generate CIM/XML bundles

bemwen:
## Description

The purpose of this PR is to add functionality to generate CIM/XML charge link bundles and write them to a stream.

This PR:
* Adds functionality to write CIM/XML to a stream based on available charge link data
* Hooks up the needed respository and serializer in our ChargeLinkBundleCreator
* Makes our EndDefault date more explicit in InstantExtensions and adds more functionality to support it
* Adds configuration of which market participant the GreenEnergyHub will present itself as when sending charges and charge links to the market
* Moves CIM serialization and deserialization code to the new folder structure in the infrastructure project
* Adds various needed mappers and tests
* Adds needed dependency injection

## References

* #649


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure/MarketDocument/Cim/IndustryClassificationMapperTests.cs
```diff
@@ -30,7 +31,22 @@ public class IndustryClassificationMapperTests
         public void Map_WhenGivenInput_MapsToCorrectEnum(string unit, IndustryClassification expected)
         {
             var actual = IndustryClassificationMapper.Map(unit);
-            Assert.Equal(actual, expected);
+            Assert.Equal(expected, actual);
```
Mech0z, line 34: Consider using Fluent assertion as we use that most of the places with actual.Should.Be(expected)

prtandrup (approved):
LGTM 🏎️

Issue #691, opened 2021-10-11
Add domain test of post office peek


BjarkeMeier:
** Synopsis **

As team Volt
We want good confidence that the charges domain always support actor peek of messages for created charge links
So that we always have a releasable domain/product

Please note that this is the first domain (black box) test of the charges domain. So it's expected to be a first-mover story fixing all the bumps on the road.

** Tech Notes **

It's all moving parts, but as of this writing the following is anticipated. Consider it helpful input rather than requirements.

- Post Office NuGet is not to be mocked, but queue name must be configurable (as discussed on dev session 11th of October)
- Post office functionality must be mocked. We assume the simplest way is to add a test-only function, that subscribes to the post office dataavailable queue and sends a peek request to the charges domain - Bjarke: This is probably not correct, I think we only would need the test function for our personal sandboxes if desired. Domain tests should not require or use such a function
- Dan recommends that:
Og jeg har tænkt mere over brug af ServiecBusListenerMock (Bjarke og jeg talte om det i fredags). Jeg er kommet frem til at det pt. er bedst at vi har en listener per eksterne queue/topic (dvs. dem jeres domæne sender til) som vi ønsker at "verify/asserte" på. Dvs. det du var i gang med Bjarke. 😊
Men i må ikke sætte listener op til at fange beskeder på jeres interne queues/topics, for så spiser de beskederne fra jeres functions. 
- Team has a agreed to put domain tests in separate test project
- We should probably use this story to move all db/repository integration tests to the integration test project

** Acceptance-Criteria **

AC 1:
Given the charges domain
When a (CIM XML) charge link creation message is received
And an actor eventually peeks
Then a bundle is sent to the Message Hub

AC 2:
Given the developer localhost setup documentation
When a developer wants to execute all XunitTest including the domain tests
He/she is able to do this by following the documentation

**Tasks**

- ~~[ ] Add setup documentation~~ (solved by @HenrikSommer)
- [x] Have Titans create test/simulation NuGet package for MessageHub behavior - [story](https://app.zenhub.com/workspaces/titans-60d029726fc7570010362aa7/issues/energinet-datahub/geh-post-office/336) created on their board
- ~~[ ] Create separate domain test project~~
- [x] Change from dataavailable to peek test
- ~~[ ] Move repository integration tests to integration test project~~ Superseeded by clean up structure task
- ~~[ ] Use UnitTest|IntegrationTest|DomainTest attributes?~~ Deemed not now due to uncertainty
[...]

PR #761, opened 2021-10-28
Add peek domain test

BjarkeMeier:
## Description

There has been some discussion in the team about whether data-available and peek should be one or two domain tests. I went with one for now as that seems to give the best overall team confidence at the moment.

The test depends on MessageHub domain-internal stuff from their NuGet package, which I expect to be subject to change. So the plan is to work with team Titans in order to have them expose such mock behavior so we can depend on officially supported functionality.

## References

* #691


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Fixtures/ChargesFunctionAppFixture.cs
```diff
@@ -179,19 +191,13 @@ protected override async Task OnDisposeFunctionAppDependenciesAsync()
...
-            {
-                if (BundleRequestQueue != null)
-                    await ServiceBusManager.DeleteQueueAsync(BundleRequestQueue.Name);
-
-                if (BundleReplyQueue != null)
-                    await ServiceBusManager.DeleteQueueAsync(BundleReplyQueue.Name);
-
-                await ServiceBusManager.DisposeAsync();
-            }
+            await ServiceBusManager.DeleteQueueAsync(BundleRequestQueue.Name);
```
lasrinnil, line 198: can the ServiceBusManager no longer be null?

lasrinnil (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Fixtures/ChargesFunctionAppFixture.cs
BjarkeMeier: Yes and no. It's test code and it _should_ never be null here. I just revert to the general approach in this class.

PR #794, opened 2021-11-05
fix: Correlation ID lost across function boundaries

bemwen:
## Description

The purpose of this PR is to make it possible for us to rely on correlation ID staying the same across function jumps.

To do this, a new middleware has been added to make sure application insight tracks the scope.

## References

* #


bemwen:
Combined with correlation ID PR from @BjarkeMeier, closed

PR #824, opened 2021-11-11
Re-enable integration tests in pipeline

HenrikSommer:
## Description

Set up keyvault url to new development environment to adapt to new integration test resource management introduce with latest Energinet.DataHub.Core.TestCommon package version.


.github/workflows/clients-ci.yml
```diff
@@ -26,20 +26,11 @@ env:
...
-  SUBSCRIPTION_ID: ${{ secrets.INTEGRATIONTEST_SUBSCRIPTION_ID }}
-  RESOURCE_GROUP_NAME: ${{ secrets.INTEGRATIONTEST_RESOURCE_GROUP_NAME }}
-  DEFAULT_LOCATION: 'westeurope'
-  TENANT_ID: ${{ secrets.INTEGRATIONTEST_TENANT_ID }}
-  CLIENT_ID: ${{ secrets.INTEGRATIONTEST_CLIENT_ID }}
-  CLIENT_SECRET: ${{ secrets.INTEGRATIONTEST_CLIENT_SECRET }}
   AZURE_KEYVAULT_URL: ${{ secrets.AZURE_KEYVAULT_URL }}
   # Variables necessary to start function for tests (similar to local use of 'local.settings.json')
-  APPINSIGHTS_INSTRUMENTATIONKEY: ${{ secrets.INTEGRATIONTEST_APPINSIGHTS_INSTRUMENTATIONKEY }}
+  APPINSIGHTS_INSTRUMENTATIONKEY: ${{ secrets.AZURE_APPI_INSTRUMENTATION_KEY }}
```
BjarkeMeier, line 31: Why does this differ from the other ("INTEGRATIONTEST_...") secrets?

.github/workflows/dotnet-solution-ci.yml
```diff
@@ -79,9 +76,12 @@ jobs:
...
 
-      - name: Test
-        shell: bash
-        run: dotnet test --configuration Release ${{ inputs.SOLUTION_FILE_PATH }}
+      - name: Run tests
+        run: |
+          az login --service-principal --username "${{ secrets.AZURE_SPN_ID }}" --password "${{ secrets.AZURE_SPN_SECRET }}" --tenant "${{ secrets.AZURE_TENANT_ID }}"
+          az account set --subscription "${{ secrets.AZURE_SUBSCRIPTION_ID }}"
+          dotnet tool install --tool-path ./temp/reportgenerator dotnet-reportgenerator-globaltool
+          dotnet test ${{ inputs.SOLUTION_FILE_PATH }} --configuration Release --verbosity normal --collect:"XPlat Code Coverage" -- DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.Format=opencover --output ${{ github.workspace }}/output
```
BjarkeMeier, line 84: Looks like a candidate for an action - probably a shared one.

.github/workflows/clients-ci.yml
HenrikSommer: Actually, I think all of those with prefix `INTEGRATIONTEST_` can be removed.

.github/workflows/dotnet-solution-ci.yml
```diff
@@ -20,6 +20,8 @@ on:
         required: true
         type: string
     secrets:
+      AZURE_KEYVAULT_URL:
```
BjarkeMeier, line 23: MD, why is this file a workflow instead of an action?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Fixtures/ChargesFunctionAppFixture.cs
```diff
@@ -80,6 +88,8 @@ protected override void OnConfigureEnvironment()
             // * DOMAINEVENT_SENDER_CONNECTION_STRING
             //
             // All other settings are overwritten somewhere within this class.
+            // Environment.SetEnvironmentVariable("INTERNAL_SERVICEBUS_RETRY_COUNT", "3");
```
BjarkeMeier, line 91: Remove?

BjarkeMeier (approved):

.github/workflows/dotnet-solution-ci.yml
Renetnielsen: If im interpreting this question correctly as to why we used a reusable workflow instead of an action?
There are several reasons, but one of the important ones is that we can then control the running OS, environmental settings. You can't do this with actions.
There is generally alot of thing you can't do with actions, and alot of stuff that makes it easier when using a workflow.

In regards to this SPECIFIC file. It should be deleted.
@HenrikSommer As far as i can see this file is not needed anymore, since you are using the shared flow.

Renetnielsen (changes requested):

.github/workflows/dotnet-solution-ci.yml
HenrikSommer: You are right, @Renetnielsen, we reference the shared workflow in the GEH .github repo instead. The file is deleted with af67786a

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Fixtures/ChargesFunctionAppFixture.cs
HenrikSommer: Removed with 72cfe9ab

.github/workflows/dotnet-solution-ci.yml
HenrikSommer: I will discuss this with @dstsyst

Renetnielsen (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Fixtures/ChargesFunctionAppFixture.cs
```diff
@@ -36,6 +36,7 @@ public class ChargesFunctionAppFixture : FunctionAppFixture
         public ChargesFunctionAppFixture()
         {
             AzuriteManager = new AzuriteManager();
+            IntegrationTestConfiguration = new IntegrationTestConfiguration();
             DatabaseManager = new ChargesDatabaseManager();
 
             var integrationTestConfiguration = new IntegrationTestConfiguration();
```
dstsyst, line 42: You should delete this line; currently you create two instances of IntegrationTestConfiguration.

Also don't call "new TestDiagnosticsLogger()", the base class has an instance that you can use (TestLogger).

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/Fixtures/ChargesFunctionAppFixture.cs
```diff
@@ -64,22 +65,29 @@ public ChargesFunctionAppFixture()
...
+                return;
+
+            var buildConfiguration = GetBuildConfiguration();
+            hostSettings.FunctionApplicationPath = $"..\\..\\..\\..\\GreenEnergyHub.Charges.FunctionHost\\bin\\{buildConfiguration}\\net5.0";
         }
 
         /// <inheritdoc/>
         protected override void OnConfigureEnvironment()
         {
             // NOTICE:
```
dstsyst, line 85: This NOTICE is not correct anymore and should be deleted; we do not set specific settings in YML anymore.

dstsyst:
Added a few comments >(

Issue #792, opened 2021-11-05
Deliver mocked charge links data to Frontend for a single (consumption) metering point


PerTHenriksen:
**Synopsis:**
Upon searching and viewing a single consumption metering point in a Graphical User Interface (GUI), the Charges domain must be able to provide charge links data associated with that metering point (a viewer could be a Grid Access Provider, an Electricity Supplier, Support line, etc.)

A GUI user is expected to search and view a single metering and within that screen a simplistic overview of the charge links associated with the given metering point are displayed.

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/b3c2d7d8-c447-403f-82c9-28bfc826e664)

**Tech notes**
* Volt: Expose API to BFF (Backend For Frontend)
* Volt: Expose API to Frontend from BFF, accessible through swagger UI
* The charge links data served will be mocked. A later story will link actual charge links data from storage to frontend (#827)
* The data served to the frontend is documented
* Thomas from team Batman has already made a similar solution for GetMeteringPointByGsrn, example can be found [here](https://github.com/Energinet-DataHub/greenforce-frontend/tree/main/apps/dh/api-dh)
* For any BFF assistance, please reach out to Dan (Mighty Ducks)
* Alexander (Mighty Ducks) is the initial consumer of this.

**Out of scope:**
* Charge Links for other types of Metering Points than consumption
* Auth / data access rights to charge links data
* Pagination support is not required as the max amount of charge links per type per metering point is expected to be 15 (rare cases).

**Acceptance criteria:**

1.  - [ ] 
Given a consumption metering point
When a request for charge links data for that specific metering point is made (at the BFF)
Then the response contains all the mocked charge links for that given metering point
<br>

2.  - [ ]
Given a valid request for charge links data is processed
When responding
The response contains the following data for a charge link:

* Charge Type
* Charge Id (aka SenderProvidedChargeId)
* Charge Name
* Charge Owner Id
* Charge Owner Name
* Tax or TaxIndicator (Only relevant for Tariff links)
* TransparentInvoicing
* Factor (or Quantity) - Only relevant for Fee and Subscription links)
* StartDateTimeUtc (or Valid From)
* EndDateTimeUtc (or Valid To) - OBS: if set to max/infinite in the data model - we will return null)

3.  - [ ] 
Given a consumption metering point with ZERO charge links
When a request is made
Then the response returned is 404 not found

<hr>

**Task list:**
_Backend work:_ 
[...]

PR #848, opened 2021-11-18
Add Web API that delivers mocked charge links data for a metering point

prtandrup:
## Description
This is the first PR of several PRs in order to deliver charge link data for a metering point to the frontend (through a BackendForFrontend (BFF)).

With this PR a Charges Web API project is introduced, which has the ability to deliver mock charge links data for a single metering point on request. Note, when metering point id input is 404, it returns "Not Found".

Later PRs will 
- introduce a Client (to be used by the BFF) to consume the backend Web API.
- use the client in the BFF (done in the greenforce-frontend repo)

A later story will replace the mock data with actual data from storage.

## References

* #792


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/Models/ChargeLinkDto.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    public class ChargeLinkDto
+    {
+        public ChargeType ChargeType { get; set; }
+
+        public string ChargeId { get; set; }
+
+        public string ChargeName { get; set; }
+
+        public string ChargeOwnerGln { get; set; }
```
BjarkeMeier, line 28: Though it's a GLN number, should it be ChargeOwnerId to align with other naming?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/Startup.cs
```diff
@@ -0,0 +1,72 @@
...
+                app.UseDeveloperExceptionPage();
+                app.UseSwagger();
+                app.UseSwaggerUI(c => c.SwaggerEndpoint("/swagger/v1/swagger.json", "GreenEnergyHub.Charges.WebApi v1"));
+            }
+
+            app.UseHttpsRedirection();
+
+            app.UseRouting();
+
+            app.UseAuthorization();
```
BjarkeMeier, line 64: Is this used or needed? Will it ever be?

BjarkeMeier (approved):
Looks great Peter.

One minor thing, when I run it on localhost I get a security warning, which I have to circumvent (in Chrome). Would be nice with documentation about (1) that there is swagger and (2) how to solve the security warning. Perhaps a short README.md in the root of the project?
image

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/Models/ChargeLinkDto.cs
prtandrup: Changed, 655146bf

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/Startup.cs
prtandrup: For now it isn't and I've removed it. Commit 7de6432d

prtandrup:
> Looks great Peter.
> 
> One minor thing, when I run it on localhost I get a security warning, which I have to circumvent (in Chrome). Would be nice with documentation about (1) that there is swagger and (2) how to solve the security warning. Perhaps a short README.md in the root of the project? image

👍 I think you are missing an IIS Express self-signed certificate. For Rider please try this:
https://blog.jetbrains.com/dotnet/2019/03/19/running-debugging-asp-net-core-apps-iis-express-rider/#:~:text=Generating%20a%20self%2Dsigned%20SSL%20certificate

lasrinnil (approved):

.github/workflows/ci.yml
```diff
@@ -50,6 +50,14 @@ jobs:
       DOTNET_VERSION: '5.0.202'
       ARTIFACT_NAME: functionhost
 
+  webapi_ci:
```
Renetnielsen, line 53: Should be dependent on solution_ci also, just to make sure we dont make unessecary artifacts if tests is not passing.

I can see the others dont either, feel free to fix that in this PR also. (Functionhost_ci and database_ci)

Renetnielsen (changes requested):

.github/workflows/ci.yml
prtandrup: Thx for input @Renetnielsen, I've added the missing dotnet_solution_ci to webapi_ci, functionhost_ci and database_migration_ci.

Renetnielsen (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/Models/ChargeLinkDto.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public bool TaxIndicator { get; set; }
+
+        public bool TransparentInvoicing { get; set; }
+
+        public int Factor { get; set; }
+
+        public DateTime StartDateTime { get; set; }
+
+        public DateTime? EndDateTime { get; set; }
```
BjarkeMeier, line 40: Should these be `Instant`s?

.github/workflows/ci.yml
BjarkeMeier: @Renetnielsen,
- Why should functionhost_ci depend on ci_base and terraform_validate?
- The name ci_base.yml doesn't tell what it does (markdown and license check)
- The name `dotnet-create-function-artifact.yml` seems misleading. It's especially surprising (judged by the name) that it is used for the DbUp project
- I believe that it's rather our `create_prerelease` job that should depend on ci_base and terraform_validate(?)

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/Models/ChargeLinkDto.cs
prtandrup: I think the use of DateTime is correct as this data is going out of the domain, not passed around internally.

Side note: I expect the ´ChargeLinkDto´ will be moved to the coming package ´Energinet.DataHub.Charges.Clients.Abstractions´ in my next PR and then used by Controller in this Web API.

.github/workflows/ci.yml
Renetnielsen: @BjarkeMeier 

> - Why should our job functionhost_ci depend on ci_base and terraform_validate?

It makes no sense to start building artifacts if we dont even have the licenses, markdown stuff in place

> - The name ci_base.yml doesn't tell what it does (markdown and license check)

The point of this file is to do all the BASE stuff that is required for a pull request to go through at any time, this includes license checks, markdown check and whatever.
Its meant to be a file that we from a central location can expand to always include the "always-needed-checks". (Meaning we at some point will fix this to `@main` instead)

If we had to reflect this in the name like markdown_license_x_x_x_check.yml we would have to make pull request to all repos everytime a change is introduced in the file. We do not want that

> The name dotnet-create-function-artifact.yml seems misleading. It's especially surprising (judged by the name) that it is used for the DbUp project

We are aware of this, we will change it to something along the lines of `dotnet-create-artifact` instead

> I believe that it's rather our create_prerelease job (rather than functionhost_ci) that should depend on ci_base and terraform_validate(?)

Answered this above in the other question

BjarkeMeier: Thx @Renetnielsen. Makes sense 👍 
Btw, it would be nice if your answer was reflected as description or similar in the shared workflows/actions. Then I could have read the intent by myself (I actually did look into them before posting my questions 😉)

Issue #823, opened 2021-11-10
Enable Grid Access Providers to peek confirmation messages for valid Charges Links they've submitted


prtandrup:
**Synopsis:**
As a Grid Access Provider that has submitted a valid `Charge Link` message
I want the ability to get (peek) a confirmation message for it
So that I know the GreenEnergyHub has (processed and) accepted it

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/c8c223ac-a0dc-41c0-b89a-26497c78cf8e)

**Tech notes:**
RSM-030 CIM XML confirmation message name: ConfirmRequestChangeBillingMasterData
Rejection messages are out of scope.
It is expected that this story includes as a minimum the following unit, integration and domain tests:
- Unit tests: Create CIM XML bundle for confirmation for `ConfirmRequestChangeBillingMasterData`
- Integration test: Write/Read from the new data available table that will be created as part of this story.
- Domain test: Submit Charge Link -> Asserts: DataAvailableNotification (on CorrelationId) & Confirmation message (on correlationId if possible, otherwise on some identifying content)

**Acceptance criteria:**
**AC1** 
Given a valid Charge Link is submitted by a Grid Access Provider, 
When the Grid Access Provider peeks for messages
Then a CIM XML confirmation message (RSM-030) is provided with correct content informing that the submitted Charge Link was accepted.

bemwen:
Note: Also added support for RSM-033 confirmation and rejection on top of the RSM-030 confirmation.

prtandrup:
#### Tested in U-002
Test passed ✅ 

Submitted new Charge Link ('ChargeId04' for MP ID: 571313180000000005, as Grid Access Provider: 8100000000023)

Postman:
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/7b368a23-fd2e-4f5e-8fe9-0b5e68af097a)

U-002 database:
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/5fea30ad-5e3c-4169-b2b9-1dbe828d7cf1)

Grid Access Provider peeks:
````There are some challenges with the MessageHub, so peeking is not a straight forward thing to do````
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/b9fc0f94-c8df-4b36-8b38-dbae5ebfd56b)
And content is as expected.


[...]

PR #900, opened 2021-11-30
refac: MessageHub bundles

bemwen:
## Description

This is only step one of the refactoring of how we store, notify and bundle messages to the mesage hub. More PRs will come, but this has been made into a single PR for ease of reading and reviewing.

This PR address bundling. Notification and message type will be addressed later.

This PR:
* Introduces a generic IAvailableDataRepository interface to replace old specific interfaces
* Introduces a generic ICimSerializer interface to replace old specific interfaces
* Replaces all BundleCreator implementations with a single implementation that makes use of the generic nature of our bundling
* Fixes AvailableDataBase constructors to be proctected as originally intended which ended up public by mistake

## References

* #823


BjarkeMeier (approved):
Great work! Looking forward to the next upcoming refactorings.

I don't know if you have considered it but I think we should really consider segregating MessageHub stuff into its own solution/microservice.

PR #908, opened 2021-12-02
refac: MessageHub notifications

bemwen:
## Description

This PR is part of the ongoing refactoring of our interactions with the MessageHub. 

The purpose of this PR is to rework the way we use notifications to the MessageHub by identifying the multiple responsibilities that was in the notifiers before and extract some of these to their own separate classes which will make them behave more alike.

Unfortunately, due the nature of the PR, alot of code is shifted around between classes which makes file comparison guess wrong. Sorry for this inconvenience to the reviewers.

This PR:
* Introduces a generic interface and implementation for notifying
* Introduces a generic interface and implementation of creating notification data
* Extracts responsibility of creating notification DTOs to a new generic notification factory
* Introduces a generic interface for creating available data
* Extracts responsibility of determining how many available data are needed into the available data factories
* Introduces a generic interface for specifying how to bundle
* Adds bundle specifications for existing message types
* Removes old notifiers which is now handled by the generic class instead
* Adds various tests previously missing

## References

* #823


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MessageHub/AvailableDataNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            _availableDataFactory = availableDataFactory;
+            _availableDataRepository = availableDataRepository;
+            _availableDataNotificationFactory = availableDataNotificationFactory;
+            _dataAvailableNotificationSender = dataAvailableNotificationSender;
+            _correlationContext = correlationContext;
+            _bundleSpecification = bundleSpecification;
+        }
+
+        public async Task NotifyAsync(TInputType input)
```
Mech0z, line 50: Should this naming reflect that its also storing data?

bemwen: I am a bit torn on that. I can see why you would want it.

On the other hand, the fact that we need to store is curiosity of how the interface between the MessageHub and the domains are made that I would not want to leak out of the interface/class.

I can see it from both perspectives.

Mech0z (approved):
Looks good, only thing is I think we should consider if the database should get similar treatment with all the common data all AvailableData share between them

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/MessageHub/AvailableChargeDataFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            _marketParticipantRepository = marketParticipantRepository;
+            _messageMetaDataContext = messageMetaDataContext;
+        }
+
+        public async Task<IReadOnlyList<AvailableChargeData>> CreateAsync(ChargeCommandAcceptedEvent input)
+        {
+            var result = new List<AvailableChargeData>();
+
+            if (input.Command.ChargeOperation.TaxIndicator)
```
lasrinnil, line 46: Could a method name describe why if the tax indicator is true, we do something? Why only if it is true?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/ChargeLinks/MessageHub/AvailableChargeLinksDataFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+            var result = new List<AvailableChargeLinksData>();
+
+            foreach (var link in acceptedEvent.ChargeLinksCommand.ChargeLinks)
+            {
+                var charge = await _chargeRepository.GetAsync(new ChargeIdentifier(
+                    link.SenderProvidedChargeId,
+                    link.ChargeOwner,
+                    link.ChargeType)).ConfigureAwait(false);
+
+                if (charge.TaxIndicator)
```
lasrinnil, line 61: Could a method name describe why if the tax indicator is true, we do something? Why only if it is true?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MessageHub/AvailableDataNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            await _availableDataRepository.StoreAsync(availableData).ConfigureAwait(false);
+        }
+
+        private async Task NotifyMessageHubOfAvailableDataAsync(IReadOnlyList<TAvailableData> availableData)
+        {
+            var notifications = CreateNotifications(availableData);
+
+            var dataAvailableNotificationSenderTasks = notifications
+                .Select(x => _dataAvailableNotificationSender.SendAsync(_correlationContext.Id, x));
```
lasrinnil, line 77: x? what is x?

lasrinnil (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/MessageHub/AvailableChargeDataFactory.cs
bemwen: Added method and comment in method

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/ChargeLinks/MessageHub/AvailableChargeLinksDataFactory.cs
bemwen: Added method and comment in method

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MessageHub/AvailableDataNotifier.cs
bemwen: Changed to be more readable

PR #912, opened 2021-12-04
feat: Add charge rejection mesage hub support

bemwen:
## Description

The purpose of this PR is to add support for charge rejections in the message hub flow.

This PR:
* Adds bundle specification for rejections
* Adds data available factory for rejections
* Renames rejection endpoint to match new naming scheme
* Adds needed dependency injection
* Cleans up a bunch of classes and contracts no longer needed (more cleanup to come later)

## References

* #823


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/MessageHub/AvailableChargeRejectionDataFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+                    Guid.NewGuid(), // ID of each available piece of data must be unique
+                    ReceiptStatus.Rejected,
+                    input.Command.ChargeOperation.Id,
+                    GetReasons(input)),
+            };
+
+            return Task.FromResult(result);
+        }
+
+        private List<AvailableChargeReceiptDataReasonCode> GetReasons(ChargeCommandRejectedEvent input)
```
lasrinnil, line 53: Does this method really belong to a factory class?

bemwen: It is part of the mapping between events and available data, so I don see why it would not belong here? Anything specific you are thinking of?

lasrinnil (approved):

Issue #886, opened 2021-11-26
Operation ID for charge links is lost


BjarkeMeier:
This property should be provided by the sender:
![image.png](https://images.zenhubusercontent.com/60c2e712ae37899de9483e3a/4baa07c5-209d-4409-a5fa-0ffed3187e13)

But apparently it's not:
![image.png](https://images.zenhubusercontent.com/60c2e712ae37899de9483e3a/dfc72f74-9154-4394-9198-08583adcbffd)

bemwen:
It turns out this bug wasnt an issue. Rather, it confused because of the naming of the class.

ChargeLinksCommandFactory is only used when creating Default Charge Links. A command is not used as part of the normal BRS-037 flow as the command is the direct output of a protobuf contract.

So, long story short, the code works.

I am opted to rename the factory as a "fix" to this bug as it was clearly causing confusion.

PR #942, opened 2021-12-10
comment on why the TSO sets OperationId when creating def. links

lasrinnil:
## Description

This PR will clarify when the TSO creates default charge links, a new operationId is supplied by the system, instead of the usual, with the actors providing the operationIds.

## References

* #886


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeLinksCommands/DefaultChargeLinksRequestCommandFactory.cs
```diff
@@ -81,7 +81,7 @@ public class DefaultChargeLinksRequestCommandFactory : IDefaultChargeLinksReques
                     EndDateTime = pair.Key.EndDateTime,
                     ChargeOwner = pair.Value.Owner,
                     StartDateTime = pair.Key.GetStartDateTime(meteringPoint.EffectiveDate),
-                    OperationId = Guid.NewGuid().ToString(), // TODO: Fix and add unit test
+                    OperationId = Guid.NewGuid().ToString(), // The TSO creates these links, a unique id is provided.
```
BjarkeMeier, line 84: Could we also update the documentation of the ChargeLinkDto.OperationId prop?

lasrinnil: done.

BjarkeMeier (commented):
There is something about this concept, design and naming...
Can we talk?

BjarkeMeier (approved):

Issue #937, opened 2021-12-09
VR: Charge update and stop requests are rejected via new validation rule


bemwen:
**Synopsis**
Given that functionality for updating and stopping a **charge** is not available when we let external parties start testing, we've decided to implement a temporarily validation rule that rejects any update or stopping requests. This is considered the most 'clean' way of handling this to avoid giving external parties a false idea that their updates and stops were processed.

As it is not possible to tell whether a Charge CIM XML is a 'create' or 'update', we need to check if the charge exists in data storage in order to determine that. 

**Acceptance Criteria:**
AC 1 Update and stops are not supported, and will be handled by an added validation rule, with an error text reflecting this.
AC 2: New validation rule to be added to general validation rule excel sheet
AC 3: Overview of active validation rules to be added to github 

**Notes: **
Discussion notes from PI planning:
Given a charge was created (with or without prices)
When any prices are submitted afterwards
This rule will reject it. (agreed, candidate for change after actor test start)

bemwen:
AC1: Validation rule added, but without text so far, as we do not yet have that functionality. It will be added in #922 instead.
AC2: Added to validation master as follows

![image.png](https://images.zenhubusercontent.com/60bdd871726f411a853d94a9/62abf9ff-699e-4289-9697-7fb39d920fe4)

PR #952, opened 2021-12-13
feat: Add validation stopping updates to an existing charge

bemwen:
## Description

The purpose of this PR is to add a validation rule which fails requests about a charge that exist in the database.

This will make sure that market participants gets a rejection until we actually implement updates and stops.

This PR:

* Adds business validation rule failing if the charge exist
* Adds the validation rule to the validation rule factory
* Adds and updates tests

## References

* #937

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/BusinessValidation/ValidationRules/ChargeUpdateNotYetSupportedRule.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/BusinessValidation/ValidationRules/ChargeUpdateNotYetSupportedRule.cs
@@ -1,0 +1,32 @@
+﻿// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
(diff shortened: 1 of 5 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/BusinessValidation/ValidationRules/ChargeUpdateNotYetSupportedRule.cs
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GreenEnergyHub.Charges.Domain.Charges;
+
+namespace GreenEnergyHub.Charges.Domain.Dtos.ChargeCommands.Validation.BusinessValidation.ValidationRules
+{
```
prtandrup, line 18: Perhaps we should consider a comment that makes it explicit that this rule covers both charge updates and stops.

prtandrup (approved):
LGTM with a minor comment

PR #961, opened 2021-12-14
Fix: ChargeLinkClient was not deserializing correctly

prtandrup:
## Description

There was a deserializing issue in the ChargeLinkClient where ChargeLinkDto was not deserialized correctly; no values were set.

This PR fixes the issue by using `new JsonSerializerOptions(JsonSerializerDefaults.Web)`

Edited an existing unit test to cover this issue.

## References

* #792


bemwen (approved):

BjarkeMeier (commented):
Good, but we should add a unit test to prevent this problem from re-occuring.

HenrikSommer (approved):

prtandrup:
> Good, but we should add a unit test to prevent this problem from re-occuring.

Yep, edited an existing unit test to cover [this](https://github.com/Energinet-DataHub/geh-charges/pull/961/commits/713508b33a988d3c56bfbfdb0c07bbe28fad8948).

PR #963, opened 2021-12-15
doc: Add charge validation documentation

bemwen:
## Description

The purpose of this PR is to add documentation of validation rules that currently exist in the charge domain.

This PR:
* Adds documentation of charge validation rules
* Updates unknown VR rule to be VR903 as per new information

## References

* #937


Mech0z (approved):

BjarkeMeier (commented):
How is the reader supposed to know what a VR rule is? Is there any doc we can reference? Or should we add som additional info or...?

PR #968, opened 2021-12-16
feat: Add errorflow support when speaking with message hub

bemwen:
## Description

The purpose of this PR is to add support for notifying the message hub of errors instead of just failing and letting the message hub wait in suspense.

Hopefully this will help on two of our current issues:
* That we sometimes get errors in our CI which is not actual there
* That we get wiser about why the error in the CI and errors in the environments occur (by logging the information)

This PR:
* Adds new exception thrown when we find no available data on the IDs requested by the message hub
* Catches exceptions in the bundle sender so that it can reply with errors to the message hub (which then does not have to wait for time out)
* Logs information about errors so we can get wiser on the current issues
* Replies to the message hub with two different kind of errors depending on situation

## References

* #0000


prtandrup (approved):
LGTM

Issue #976, opened 2021-12-20
Use class instead of record in Charges API


ZavezX:
### Description
We should use class instead of record for return object as we need to annotate with required attribute in order to generate typescript model with non-nullable properties.

### Acceptance Criteria
- [ ] Class is used instead of record
- [ ] Package version is increased
- [ ] Release notes are made

PR #978, opened 2021-12-20
Use class instead of record in api

ZavezX:
## Description

- Use class instead of record and annotate properties with `Required` where value cannot be null in order to tell swagger to add property to required section for the frontend typescript generation to function properly.

## References

* #976


source/Energinet.Charges.Libraries/source/Energinet.DataHub.Charges.Clients/documents/release-notes/release-notes.md
```diff
@@ -1,5 +1,9 @@
 # Energinet.DataHub.Charges.Clients Release notes
 
+## Version 1.0.20
+
+Use class for `ChargeLinkDto` insted of record.
```
ZavezX, line 5:
```suggestion
Use class for `ChargeLinkDto` instead of record.
```

BjarkeMeier (commented):
Please preserve the props documentation.

Issue #946, opened 2021-12-10
Tech: Segregate MessageHub into it's own stack


BjarkeMeier:
**Synopsis**

Improve maintainability of the system by segregating the concerns of MessageHub communication from core business.

**Acceptance-Criteria**

AC1
Application, Domain and Infrastructure projects no longer holds any MessageHub related code.

AC2
MessageHub code has a design and and architecture that is suitable for this kind of software.

**Tech Notes**

This story was selected by the dev team for the upcoming sprint 7.1.

Please note that the solution might imply introducing a new event to support the reply in BRS-004.
The dev team decided to host in the existing function host.
Michael suggested that the number of function endpoints could be even more reduced.

bemwen:
*AC1:*
Only found IHubSenderConfiguration i Domain and its implementation in infrastructure which appear to be message hub specific.

Otherwise I found no trace of the message hub in Application, Domain or Infrastructure

*AC2:*
MessageHub design and architecture has mostly been left as designed, with the exception of splitting BRS-004 flow a bit. Rather, the code has been moved around to be easier to read and understand.

It seems appropriate for the software use.

BjarkeMeier:
The hub sender configuration as per Michael's test comment is being fixed in PR #986.

BjarkeMeier:
Well, it _is_ actually used in order to create e.g. recipient in `ChargeLinksCommand`. See `source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeLinksCommands/ChargeLinksCommandFactory.cs`. So I disagree with the test comment that it could/should be moved to the message hub stack.
I consider the story done.

HenrikSommer:
AC1:
I have reviewed the IHubSenderConfiguration and its implementation with the developer and I see no Message Hub specific code, hence I agree that the acceptance criterion is fulfilled.

PR #979, opened 2021-12-21
Improve naming and file/folder structure

BjarkeMeier:
Main improvement is that the message hub endpoint publishes
an event about that it has notified the Message Hub rather
that telling that charge links were created as that is no
business of the message hub stack.

## Description
Improve naming and file/folder structure

Main improvement is that the message hub endpoint publishes an event about that it has notified the Message Hub rather that  telling that charge links were created as that is no business of the message hub stack.

In particular, the infrastructure has undergone quite some refactoring for better maintainability:

image

image

## References

* #946


bemwen (approved):
LGTM

PR #986, opened 2021-12-23
Move hub sender configuration to MessageHub stack

BjarkeMeier:
## Description

Hub sender configuration is only used for MessageHub.
Found by @bemwen during test of #946.

## References

* #946


BjarkeMeier:
Well, it _is_ used in order to create e.g. recipient in `ChargeLinksCommand`. See `source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeLinksCommands/ChargeLinksCommandFactory.cs`.

PR #1027, opened 2022-01-12
Restore SonarCloud check

Sondergaard:
## Description

This PR will restore SonarCloud workflow in the build pipeline. The reason for this is that we need to be ready for a security audit in near future.

At a later stage this will be moved to be a part of the generic build flow in geh-core.


.github/workflows/sonarcloud-check.yml
```diff
@@ -0,0 +1,58 @@
+name: SonarCloud Check
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - source/**
+      - .github/workflows/sonarcloud.yml
```
BjarkeMeier, line 8:
```suggestion
      - .github/workflows/sonarcloud-check.yml
```

or rename the file and fix the other trigger.

.github/workflows/sonarcloud-check.yml
```diff
@@ -0,0 +1,58 @@
...
+    types: [opened, synchronize, reopened]
+    paths:
+      - source/**
+      - .github/workflows/sonarcloud-check.yml
+
+env:
+  SOLUTION_PATH: 'source\GreenEnergyHub.Charges\GreenEnergyHub.Charges.sln'
+  DOTNET_VERSION: '5.0.202'
+  SONAR_PROJECTKEY: 'geh-charges'
+  SOURCE_PATH_ON_BUILDMACHINE: 'D:\a\geh-charges\geh-charges\source' # We have to be speficic with this path to get around a .NET 5.0 issue in SonarCLoud
```
BjarkeMeier, line 20: Are there any possible problems here regarding parallel PR executions?

BjarkeMeier (commented):
Should we add this to the PR gate?

.github/workflows/sonarcloud-check.yml
Sondergaard: I have only restored the previous configuration. That file contained `-check` and the configuration was updated to reflect this. 

What would you like the end result to be? I will change accordingly.

.github/workflows/sonarcloud-check.yml
Sondergaard: I really don't know - again I restored the previous configuration.

I don't think that github would schedule two runs for the same repository to the same worker - but I might be wrong.

Sondergaard:
> Should we add this to the PR gate?

This will be moved at a later stage to the PR gate. For now the focus is getting it re-implemented and ready for when we have to do the inspection.

.github/workflows/sonarcloud-check.yml
BjarkeMeier: I think "check" informs us that the workflow doesn't "do" anything (like creating artifacts) but solely performs a check. So I'd go with "-check".

BjarkeMeier (approved):
I'll leave it to you whether you are going to test or investigate possible race conditions.

PR #1036, opened 2022-01-13
Charges Web API - Minor changes (remove calculated enddates and nodatime, adjust scaffolding)

prtandrup:
## Description

* Adds logic to `ChargeLinkQueryLogic` to remove any calculated `EndDefault` charge link `EndDateTime` before responding to clients. In this case, clients shall receive `"endDate": null`

* Applies `--no-onconfiguring` to scaffolding script to suppress `OnConfiguring()` method generation, which is unecessary.

* Removes `NodaTime` from `UseSqlServer` config, as it is not used in Charges Web API project. DateTime2 is used in SQL. Charges Web API responds to clients using `DateTimeOffset` in `ChargeLinkDto`.

## References

* #0000


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/IntegrationTests/WebApi/ChargeLinksControllerTests.cs
```diff
@@ -70,6 +71,21 @@ public async Task GetAsync_WhenMeteringPointIdHasChargeLinks_ReturnsOrderedCharg
...
+            var response = await _client.GetAsync($"{BaseUrl}{KnownMeteringPointId}");
+
+            // Assert
+            var jsonString = await response.Content.ReadAsStringAsync();
+            var actual = JsonSerializer.Deserialize<List<ChargeLinkDto>>(
+                jsonString,
+                GetJsonSerializerOptions());
+
+            actual.Should().NotContain(c => c.EndDate == InstantExtensions.GetEndDefault().ToDateTimeOffset());
+        }
```
BjarkeMeier, line 87: Couldn't we have tested this in a unit test?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.WebApi/ModelPredicates/ChargeLinkQueryLogic.cs
```diff
@@ -30,12 +31,12 @@ public static IQueryable<ChargeLinkDto> AsChargeLinkDto(this IQueryable<ChargeLi
...
                     c.Charge.Name,
                     c.Charge.Owner.MarketParticipantId,
-                    "Netvirksomhed XYZ", // Hardcoded as we currently don't have the data
+                    "<Aktørnavn XYZ>", // Hardcoded as we currently don't have the ChargeOwnerName data
                     c.Charge.TaxIndicator,
                     c.Charge.TransparentInvoicing,
                     c.Factor,
                     c.StartDateTime,
-                    c.EndDateTime));
+                    c.EndDateTime == InstantExtensions.GetEndDefault().ToDateTimeOffset() ? null : c.EndDateTime)); // Nullify "EndDefault" end dates
```
BjarkeMeier, line 39: I think we should unit test this.
But why don't we change it in the domain model rather than doing it here? Feels like fixing symptoms to me...

PR #1049, opened 2022-01-14
fix naming issues in tf files

Mech0z:
## References

* #0000


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.FunctionHost/Common/EnvironmentSettingNames.cs
```diff
@@ -83,9 +83,9 @@ public static class EnvironmentSettingNames
 
         // Internal, charge links, accepted
         public const string ChargeLinksAcceptedTopicName = "CHARGE_LINKS_ACCEPTED_TOPIC_NAME";
-        public const string ChargeLinksAcceptedSubConfirmationNotifier = "CHARGE_LINKS_ACCEPTED_SUB_CONFIRMATIONNOTIFIER";
```
bemwen, line 86: Remember changes in this should be aligned with ChargesServiceBusResourceNames.cs as well. Might be nice with a comment to that effect in both files?

It will work without doing it, but the domain tests will float further and further away from using the same settings as the functions themselves.

HenrikSommer (approved):

Issue #938, opened 2021-12-09
VR: ChargeLinks update and stop requests are rejected via new validation rule


bemwen:
**Synopsis**
Given that functionality for updating and stopping a **ChargeLink** is not available when we let external parties start testing, we've decided to implement a temporarily validation rule that rejects any update or stopping requests.

This validation will be checking if an incoming request's charge link period overlaps an existing charge link period (for same charge and metering point combination), if so => reject.

**Tech notes:**

* Validation rule setup has not been made for charge links yet
* Even though it is not the intention in the long run, it is currently possible to submit a charge link with an end date which will not be rejected by this rule. E.g.
Existing Charge Link with start = 01/01/2022
Same charge link submitted with start and end dates = 01/11/2021 - 01/12/2021 (accepted)
* Domain test should be updated by adding 1 failing charge link resulting in 3 message types being received by MessageHub (Rejected, Accepted and the 2 charge links)

**Tech Tasks**
~~- [ ]  (1) And why do we use ProtoBuf internally? Is it still an architect requirement?~~

~~- [ ]  (2) Only one contract. Example:    ChargeCommand = ChargeCommandReceived = ChargeCommandAccepted =     ChargeCommandRejected.ChargeCommand.~~

~~- [ ]  (3) Don't repeat enums in contracts Use int's for enum values and just cast on both sides (serialization and deserialization).~~

~~- [ ]  Make protobuf Decimal a shared type (get Søren (SA) to sign off on namespace)~~

- [x]  Ensure that any function or queue/topic name related to the Default Charge Links flow, has links in plural and 'default' included in their name, so it's clear that they are related to this sub process and may include multiple links.

- [x]  Swith to Json on internal events in Charges domain

- [x] Cleanup files used to send Protobuf internally in Charges domain

- [x] Add unit test to ensure switch on BundleType enum in BundleCreatorProvider is always complete

- [x] Create meaningful reject messages to market participants for all validation rules

**Acceptance Criteria:**
AC 1 Update and stops are not supported, and will be handled by an added validation rule, with an error text reflecting this.
AC 2: New validation rule to be added to general validation rule excel sheet and to overview on GitHub

lasrinnil:
@HenrikSommer and I have discussed the Json / Protobuf usage internally in the Charges domain. We found the existing overhead of maintaining the existing internal protbuf contracts and their mappers etc, are quite high. And would only grow in the future. So we think it would be beneficial to switch to Json internally.

[...]

PR #1052, opened 2022-01-17
Fix: Add missing subscription for rejected charge links

prtandrup:
## Description

This PR adds a missing service bus subscription to the application settings of the function host:

* `CHARGE_LINKS_REJECTED_SUBSCRIPTION_NAME` was not set in `func-functionhost.tf`

## References

* #938


HenrikSommer (approved):

BjarkeMeier (approved):
How about adding the corresponding check in the health endpoint?

PR #1062, opened 2022-01-19
feat: use JSON for internal communication

lasrinnil:
## Description

Implement communication internally between queues now uses JSON instead of Protobuf.
Remove all the leftover code which enabled sending messages in Protobuf internally.

Externally we still use Protobuf.

## References

* #938


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/DomainTests/ChargeIngestionTests.cs
```diff
@@ -32,7 +32,7 @@ namespace GreenEnergyHub.Charges.IntegrationTests.DomainTests
     [IntegrationTest]
     public class ChargeIngestionTests
     {
-        private const int SecondsToWaitForIntegrationEvents = 15;
+        private const int SecondsToWaitForIntegrationEvents = 30;
```
BjarkeMeier, line 35: Do you want to commit this value?

BjarkeMeier (approved):
Wauw!

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/DomainTests/ChargeIngestionTests.cs
lasrinnil: NO! Thank you for catching it <3

PR #1080, opened 2022-01-24
Fix incorrect subscription in `local.settings.sample.json`

prtandrup:
Fixed subscription name, so we are able to run function-host locally against an environment. 

## References

* #0000

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.FunctionHost/local.settings.sample.json
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.FunctionHost/local.settings.sample.json
@@ -32,8 +32,8 @@
     // Integration, metering point domain
     "METERING_POINT_CREATED_TOPIC_NAME": "metering-point-created",
     "METERING_POINT_CREATED_SUBSCRIPTION_NAME": "metering-point-created-sub-charges",
-    "CREATE_LINKS_REQUEST_QUEUE_NAME": "create-link-request",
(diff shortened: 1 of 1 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.FunctionHost/local.settings.sample.json
```diff
@@ -32,8 +32,8 @@
     // Integration, metering point domain
     "METERING_POINT_CREATED_TOPIC_NAME": "metering-point-created",
     "METERING_POINT_CREATED_SUBSCRIPTION_NAME": "metering-point-created-sub-charges",
-    "CREATE_LINKS_REQUEST_QUEUE_NAME": "create-links-request",
-    "CREATE_LINKS_REPLY_QUEUE_NAME": "create-links-reply",
+    "CREATE_LINKS_REQUEST_QUEUE_NAME": "create-link-request",
+    "CREATE_LINKS_REPLY_QUEUE_NAME": "create-link-reply",
```
BjarkeMeier, line 36: This doesn't look like the right fix. You are reverting the correct plural use. I guess you are looking for a change somewhere else.

prtandrup: Good point. And you are correct. Fixing this so it's plural "all the way around" requires some changes in 3 repo's which I won't prioritize just now. I'll remove this change, but keep the subscription change, so that's at least fixed :)

HenrikSommer (approved):

Issue #951, opened 2021-12-13
System test MVP


prtandrup:
**Synopsis:**
As the Volt team
We want to execute automated system tests (business process happy flows), prior to deploying new software updates to the actor test environment
So that we will be informed about problems that potentially span across the system as a whole (across domains) and may act upon it.

**Tech notes:**
* System tests are tests that run across domains, on the system as a whole, [source](https://github.com/Energinet-DataHub/green-energy-hub/blob/main/docs/test-principals.md)
* This test is categorized as System Test and built as a xunit test project
* Dan _(@dstsyst)_ from Mighty Ducks has already started this, by setting up a skeleton (see this [branch](https://github.com/Energinet-DataHub/geh-charges/tree/dstenroejl/system-tests)) and then another branch for pipeline work (private repo, please ask Peter _(@prtandrup)_ for location)
* All communication with the system under test is done through DataHub API Management
* We should move GreenEnergyHub.Charges.SystemTests/TestFiles/RequestChangeBillingMasterData.xml and EmbeddedResourceHelper to a new project shared by IntegrationTests and SystemTests projects (already done)

**Out of scope:**
* ~~As part of the MVP, team Volt just wants to be informed about failed system test, not guarding further deployment. A general concept needs to be discussed and agreed upon across teams. Hence deployment guarding is out of scope~~
* GUI tests

**Acceptance criteria**
- [x] **AC1 // The following system test is created for BRS-031 'Create subscription'**
Given a Grid Access Provider submits a 'Charge price list request' containing a new subscription
When the Grid Access Provider peeks for new messages
Then a confirmation message is received for the particular subscription.
(Note: later down the road, we will include a check for electricity supplier(s) receives notifications for this subscription to cover the entire business process flows)
<br>

- [x] **AC2 // Executed after deployment**
Given one or more Charges system tests are ready
When deployment has been carried out to the development environment (U)
Then the system tests are executed and failed results will prevent deployment to remaining environments

<hr>

````Below is considered a bonus target - do it if time allows - otherwise please create a new story for it````
<hr>

~~- [ ] **Bonus // The following system test is created for BRS-037 'Maintain Charge Links'**
Given a Grid Access Provider submits a new 'Charge link request' containing a new Charge Link (new charge + MP ID combination)
When the Grid Access Provider peeks for new messages
Then a Confirmation message is received for the particular charge link.~~ Moved to #1265

prtandrup:
[...]

PR #1083, opened 2022-01-25
Reenable System test CI for pipeline troubleshooting

HenrikSommer:
## Description

System test CI build was previously deactivated due to transient concurrency error. It is now reenabled for further investigation.

## References

* #951


BjarkeMeier (approved):
I'm curious. What's the purpose of the system test CI? Should it really block a PR in the charges domain, let alone how unstable the u-001 environment is? I believe that a system test should rather be used as an "environment deployment gate" - say run in one environment and preventing any deployment to the "next" environment. However, again would this be desirable? That would mean that one bug somewhere in the system would prevent all other bugs from being deployed to subsequent environments. So what is the actual purpose and how do you envision that we (eventually) will use it?

prtandrup:
> I'm curious. What's the purpose of the system test CI? Should it really block a PR in the charges domain, let alone how unstable the u-001 environment is? I believe that a system test should rather be used as an "environment deployment gate" - say run in one environment and preventing any deployment to the "next" environment. However, again would this be desirable? That would mean that one bug somewhere in the system would prevent all other bugs from being deployed to subsequent environments. So what is the actual purpose and how do you envision that we (eventually) will use it?

The system test CI only builds and publishes the artifact needed to run the system tests in CD. Where we want to run the tests are up to us. I figure we start with U-001 and then evaluate.

PR #1085, opened 2022-01-26
Create SQL database based on azurerm_mssql_database TF resource

djorgensendk:
This is the first of two PRs migrating the domain SQL database to the new Terraform module based on the azurerm_mssql_database resource.
This PR creates the new resources, and when going through the CD pipeline, the DbUp operations are executed upon the new database.
The second PR removes the current used SQL database and changes the pointers from Functions and/or WebApis to point to the new database created in this PR.

The above is done in agreement with members of all product teams on January 26th.


BjarkeMeier (commented):
Can you add some PR description in order to understand the purpose and value of this PR? I doesn't seem like all the new settings are used?

djorgensendk:
> Can you add some PR description in order to understand the purpose and value of this PR? I doesn't seem like all the new settings are used?

Good point. I have added a description to the PR

HenrikSommer (approved):

PR #1089, opened 2022-01-27
Remove existing SQL DB and move pointers to the new one

djorgensendk:
This is the second of two PRs migrating the domain SQL database to the new Terraform module based on the azurerm_mssql_database resource.
This PR removed the old resources and the pointers from the application infrastructure now uses the new resources.

The above is done in agreement with members of all product teams on January 26th.


BjarkeMeier (commented):
How about the web API (`app-webapi.tf`)? I think there are some important changes missing.

djorgensendk:
> How about the web API (`app-webapi.tf`)? I think there are some important changes missing.

I have changed the value of "CHARGE_DB_CONNECTION_STRING" to point to the connection string of new database. Am I missing anything else?

BjarkeMeier:
> > How about the web API (`app-webapi.tf`)? I think there are some important changes missing.
> 
> I have changed the value of "CHARGE_DB_CONNECTION_STRING" to point to the connection string of new database. Am I missing anything else?

Good question. When looking at your PRs I wondered why you chose to change the naming of the connection string environment variable names? You actually cannot do that without changing the function host endpoints too. And that will lead you directly further into our domain tests - and that's something you wouldn't wanna mess with. So as-is I believe this PR will break the domain on deployment.

BjarkeMeier (approved):

PR #1054, opened 2022-01-17
Extend deep health check with additional queues, topics and subscriptions

prtandrup:
## Description

This PR extends the existing deep health check with:
* additional queue exist checks
* topic and subscription exist check

After implementing this, the deep health check will check queues, topics and subscriptions required for both domain as well as integration events.

Note: I have been through the app settings of the function host to figure out what queues, topics and subscription needed to be added. See [here](https://github.com/Energinet-DataHub/geh-charges/blob/db8c7700daadef85f492f6226cf7df7d9708703c/build/infrastructure/main/func-functionhost.tf#L26).

## References

* #1100

PR #1094, opened 2022-01-28
Fix - misconfiguration related to downstream Charge Links Endpoints

prtandrup:
## Description

This PR fixes some misconfiguration revealed by app-insights for the subscriptions needed for 
* ChargeLinksEventPublisherEndpoint
* ChargeLinkDataAvailableNotifierEndpoint
* ChargeLinkConfirmationDataAvailableNotifierEndpoint

It is planned to [extend our deep health check](https://github.com/Energinet-DataHub/geh-charges/pull/1054), so it also covers these subscriptions going forward.

Also, we've found that the OriginalOperationId length used in two DataAvailable must be increase to hold GUID values.

## References

* #0000


BjarkeMeier (approved):
Seems like we'd have been better off actually completing that deep health check rather than spending time searching for errors in the system. How about prioritizing the finalization of that PR?

prtandrup:
> Seems like we'd have been better off actually completing that deep health check rather than spending time searching for errors in the system. How about prioritizing the finalization of that PR?

We're stumbling upon these while testing the receipt flow 😒 . The health check PR will be completed once the receipt flow is up and running.

PR #1132, opened 2022-02-04
Update Energinet .github workflow version

HenrikSommer:
## Description

Energinet .github workflow dependencies are updated to the latest version (4.0.0), except for dotnet-create-function-artifact.yml which is updated to 3.0.2, the latest version without a bug, that is present in 3.0.3 and 4.0.0.

In addition, the CA2007 reliability rule is handled in three cases where the rule wasn't earlier forced.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,13 +20,13 @@ on:
 
 jobs:
   publish_release:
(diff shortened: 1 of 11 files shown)


BjarkeMeier (approved):
Was there a reason for making this PR?

Issue #1127, opened 2022-02-03
 HTTP POST - response for successful HTTP 200 change to 202 after a request charge or charge link


IreneDataHub:
If the content complies with the applicable rules, 202 Accepted will be returned from DataHub instead of 200 as we use today.

Acc:
Accepted 202 will be returned from DataHub after receive chargelink

Acc:
Accepted 202 will be returned from DataHub after receive charge

prtandrup:
### Tested in T-001

✅ **ChargeIngestion gives 202**

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/ec5dff23-3a12-40ea-b9fe-d16c81b144b9)

✅ **ChargeLinksIngestion gives 202:**

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/3ef20145-042d-467e-9ae0-36ae4a4798d2)

PR #1133, opened 2022-02-04
Fix: Return `StatusCode 202 Accepted` instead of 200 OK - for charges and charge links requests

prtandrup:
## Description

This PR fixes a bug which made `ChargeIngestion` and `ChargeLinkIngestion` return `StatusCode 200 OK` instead of `StatusCode 202 Accepted` for schema valid requests.

The problem was that `await httpResponse.WriteAsJsonAsync(response).ConfigureAwait(false);` changed our response.statuscode from 202 to 200.

## References

* #1127

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.SystemTests/ApiManagementTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.SystemTests/ApiManagementTests.cs
@@ -53,7 +53,7 @@ namespace GreenEnergyHub.Charges.SystemTests
 
         // This shows how we can call API Management using a valid access token
         [SystemFact]
-        public async Task When_RequestApiManagementWithAccessToken_Then_ResponseIsOk()
+        public async Task When_RequestApiManagementWithAccessToken_Then_ResponseIsAccepted()
         {
             // Arrange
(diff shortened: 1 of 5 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.SystemTests/ApiManagementTests.cs
```diff
@@ -70,7 +70,7 @@ public async Task When_RequestApiManagementWithAccessToken_Then_ResponseIsOk()
             using var actualResponse = await httpClient.SendAsync(request);
 
             // Assert
-            actualResponse.StatusCode.Should().Be(HttpStatusCode.OK);
+            actualResponse.StatusCode.Should().Be(HttpStatusCode.Accepted);
```
BjarkeMeier, line 73: Perhaps change method name to end with "The_ResponseIsAccepted".

BjarkeMeier (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.SystemTests/ApiManagementTests.cs
prtandrup: Good catch, fixed.

HenrikSommer (approved):

PR #1162, opened 2022-02-17
1151 add db migration

jonasdmoeller:
## Description

Added DB upgrade as part of persist charge update feature.

## References

* #0000


prtandrup (approved):
Approved, pair-worked

PR #1170, opened 2022-03-01
How to test Web API without a real user token

dstsyst:
## Description

An example of how we can mock `IJwtTokenValidator` during test setup and configure its behavior and e.g. ensure it validates a fake token as `IsValid`. This way we can avoid having to use real user tokens in Web API tests.

## References

https://app.zenhub.com/workspaces/mighty-ducks---the-outlaws-6193fe815d79fc0011e741b1/issues/energinet-datahub/the-outlaws/210


prtandrup (approved):
Very nice, Dan.

Issue #1072, opened 2022-01-20
VR.507-1 implementation is incorrect - It block submission of master data only as well as price series longer than a day


prtandrup:
**Description**
Our current implementation of VR.507.1 `The number of prices <1> for charge <2> doesnt match period type <3>` is incorrect.

The current implementation has three shortcomings: 

1.   It throws an `ArgumentException` when `resolution = P1M`, which blocks VR.505-1 from rejecting a wrong resolution for a tariff. 

2. It does not consider the price series period to be longer than a day, which often will be the case.

3. It blocks the submission of charges with master data only.

`ChargeTypeTariffPriceCountRule.cs`

**Acceptance criteria:**
Please see  prtandrup's comment from 15-03-2022, where updated AC's are listed
~~- [ ] VR-507-1 works as expected~~ 
~~- [ ] Submit a charge message that violates VR.505-1 and get a rejection message for it~~

Request that identified this problem:
````
<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
	<cim:mRID>DocId2022-01-20T11:22:18.278Z</cim:mRID>
	<cim:type>D10</cim:type>
	<cim:process.processType>D18</cim:process.processType>
	<cim:businessSector.type>23</cim:businessSector.type>
	<cim:sender_MarketParticipant.mRID codingScheme="A10">8100000000030</cim:sender_MarketParticipant.mRID>
	<cim:sender_MarketParticipant.marketRole.type>DDM</cim:sender_MarketParticipant.marketRole.type>
	<cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
	<cim:receiver_MarketParticipant.marketRole.type>DDZ</cim:receiver_MarketParticipant.marketRole.type>
	<cim:createdDateTime>2022-01-20T12:22:18Z</cim:createdDateTime>
	<cim:MktActivityRecord>
		<cim:mRID>OpId2022-01-20T11:22:18.278Z</cim:mRID>
		<cim:ChargeGroup>
[...]

PR #1184, opened 2022-03-04
Fix: Validation Rule VR.507-1 `ChargeTypeTariffPriceCountRule` to NOT block master data only requests and not throw when `Resolution = P1M`

prtandrup:
## Description

This PR removes some shortcomings in the way VR.507-1 `ChargeTypeTariffPriceCountRule ` is currently implemented:
* It allows master data only requests through (requests with no prices)
* It doesn't throw when resolution = P1M (monthly)

But be aware that it still rejects price series longer than a day of values.

Nevertheless, I do believe we need to reconsider this validation rule; 
* is it fit for a CIM context in its current state? 
* Should it be applied on all types of charges?
* Should it only assert that the number of Points match the time interval of the price series?
* Should we perhaps add a new validation rule for restricting price series shorter than 1 day of points?

I'll note this on the story as well for further discussion. For now, this PR will bring the validation rule to a better, but not final, place. 

## References

* #1072

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeTariffPriceCountRuleTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeTariffPriceCountRuleTests.cs
@@ -31,6 +31,14 @@ namespace GreenEnergyHub.Charges.Tests.Domain.Dtos.ChargeCommands.Validation.Inp
     [UnitTest]
     public class ChargeTypeTariffPriceCountRuleTests
     {
+        [Fact]
+        public void IsValid_WhenPointsCountIsZero_IsTrue()
+        {
+            var chargeCommand = new ChargeCommandBuilder().WithChargeType(ChargeType.Tariff).Build();
+            var sut = new ChargeTypeTariffPriceCountRule(chargeCommand);
+            sut.IsValid.Should().BeTrue();
+        }
+
         [Theory]
         [InlineAutoMoqData(1, false)]
         [InlineAutoMoqData(23, false)]
@@ -53,7 +61,6 @@ namespace GreenEnergyHub.Charges.Tests.Domain.Dtos.ChargeCommands.Validation.Inp
         }
 
         [Theory]
-        [InlineAutoMoqData(0, false)]
         [InlineAutoMoqData(1, true)]
         [InlineAutoMoqData(2, false)]
         [InlineAutoMoqData(24, false)]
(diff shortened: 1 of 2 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeTariffPriceCountRuleTests.cs
```diff
@@ -53,7 +61,6 @@ public class ChargeTypeTariffPriceCountRuleTests
         }
 
         [Theory]
-        [InlineAutoMoqData(0, false)]
```
jonasdmoeller, line 56: Why remove this test for count = 0? Shouldn't it just return `true` after this update?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeTariffPriceCountRuleTests.cs
```diff
@@ -77,14 +84,34 @@ public class ChargeTypeTariffPriceCountRuleTests
...
```
jonasdmoeller, line 80: Why remove this test for count = 0? Shouldn't it just return `true` after this update?

jonasdmoeller (dismissed):

x-platformcoder (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeTariffPriceCountRuleTests.cs
HenrikSommer: That scenario is moved to the new `IsValid_WhenPointsCountIsZero_IsTrue()` test method, which is fine, I think, as it represents the new case that master data only requests are allowed.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeTariffPriceCountRuleTests.cs
HenrikSommer: Yes, it returns `true`, see my comment above.

PR #1190, opened 2022-03-10
Removed references from infrastructure to application and moved some …

jonasdmoeller:
…interfaces to application

## Description

The application layer has a reference to the infrastructure layer, which violates the principles of a clean architecture.
This PR moves some interfaces from Infrastructure.Core to Application (keeping the implementation of interfaces in Infrastructure), and removes the outward reference.

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/ChargeLinks/Handlers/ChargeLinkEventPublishHandler.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/ChargeLinks/Handlers/ChargeLinkEventPublishHandler.cs
@@ -14,9 +14,9 @@
 
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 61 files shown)


HenrikSommer (approved):

BjarkeMeier (commented):
Are you sure you wanna do this? Perhaps you'd like to rather consider renaming InfraCore.
InfraCore are stuff that has been identified as potentially reusable and thus candidate for NuGet packages. So I don't think that they (generally speaking - there might be exceptions) violates the clean architecture. At least not except for the name of the project.

jonasdmoeller:
> 

I see. All the MessagingExtension is a good candidate to be added to the actual Messaging nuget package that already exist.

Would it be better a better approach to move exceptions violating the clean architecture, as you mention, to the actual Infrastructure layer (i.e. JwtTokenMiddleware, ServiceBus and ProtoBuf stuff) and keep reusable components in this Core project??

Issue #1185, opened 2022-03-07
Market Actors requests that Response Header must contain MessageType, for easy content identification


PerTHenriksen:
**Problem**: 
Market participants would like to know the content of the peeked messages prior to parsing it - thus the header must be extended with the message type. 

**Tech notes**: 
MessageHub has been extended with a MessageType, which must contain the CIM-type of the Document, i.e. <cim:NotifyPriceList_MarketDocument... in the header of the  'MessageType': 'NotifyPriceList'

We will need to change the implementation of the AvailableDataNotificationFactory class. The MessageType can be set as a property on DataAvailableNotificationDto.

Energinet.DataHub.MessageHub.Client (version 1.9.1) has an added field in the DataAvailableNotificationDto, intended for document type - thus text is passed on to whoever peeks. So we update the NuGet package to 1.9.1 as part of this story.

**Acceptance criteria:** 
- [x] Data available notifications is enriched with MessageType
- [x] Notify TML when the correction has been done. Then Titans will enable the functionality again (currently disabled)

jonasdmoeller:
@PerTHenriksen 
Hermed verificeret at MessageType kommer med i http header på peek af charge update besked i T-001.

![image.png](https://images.zenhubusercontent.com/60dafff3568c156496e54f35/ef1024af-585c-43a3-a0b0-e1a51ce9fb53)

PerTHenriksen:
Verified that MessageType added is added to header; ![image.png](https://images.zenhubusercontent.com/60bdf7bbfa2182da7fd5cc6d/5b70eb56-f3de-4009-9016-844c79f064e3)

PR #1191, opened 2022-03-10
1185 response header must contain message type

x-platformcoder:
## Description

Added the cim document type to the message that is delivered to the messagehub after request is validated and stored

## References

* #1185


BjarkeMeier (approved):
LGTM.
Welcome @x-platformcoder - wonder where you found that user name 😉

HenrikSommer (commented):
We need to consider whether this change has an impact on any, some, or all inheritances of BundleSpecification.

PR #1192, opened 2022-03-10
1152 cancel stop charge

jonasdmoeller:
## Description

Added handling of cancel stop event.

## References

* #0000

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/Handlers/ChargeCommandReceivedEventHandler.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/Handlers/ChargeCommandReceivedEventHandler.cs
@@ -80,13 +80,13 @@ namespace GreenEnergyHub.Charges.Application.Charges.Handlers
                     await HandleCreateEventAsync(commandReceivedEvent.Command).ConfigureAwait(false);
                     break;
                 case OperationType.Update:
-                    HandleUpdateEvent(charge, commandReceivedEvent.Command);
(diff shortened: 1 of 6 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/Handlers/ChargeCommandReceivedEventHandler.cs
```diff
@@ -84,14 +80,13 @@ await _chargeCommandReceiptService
...
                 case OperationType.Update:
-                    if (charge == null)
-                        throw new InvalidOperationException("Could not update charge. Charge not found.");
                     HandleUpdateEvent(charge, commandReceivedEvent.Command);
                     break;
                 case OperationType.Stop:
-                    if (charge == null)
-                        throw new InvalidOperationException("Could not stop charge. Charge not found.");
-                    charge.Stop(commandReceivedEvent.Command.ChargeOperation.EndDateTime);
+                    charge?.Stop(commandReceivedEvent.Command.ChargeOperation.EndDateTime);
```
BjarkeMeier, line 86: Is it correct to silently ignore all operations if the charge is not found?

jonasdmoeller: No we should not continue if null. I will change it to use the null-forgiving operator - as null will always be treated as a `OperationType.Create`

This is all subject to change, so we are aware of the temporary bad design.

HenrikSommer (approved):

Issue #1154, opened 2022-02-10
Persist Charge (4/4) - Update before a registered stop flow (ac 1a)


prtandrup:
**Synopsis:**
As a Grid Access Provider,
I want to update the master data of a charge that has a future stop date
So that I can keep the master data aligned with my own system until the charge is stopped.

This is another story of a few stories with the goal of implementing the [charge persistence rule set](https://github.com/Energinet-DataHub/geh-charges/tree/main/docs/process-flows#persist-charge) (master data only)

With this story we will implement the **updating before a registered stop** flow, highlighted below:
<hr>

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/4de7377f-2396-4f74-b9a9-6fc70c918479)

**Acceptance criteria:**
- [ ] 1 Given an existing charge with two periods has been stopped
When updating the charge master data with an effective date inside period two
Then period two's end date is adjusted
And the master data update is inserted as a third period with the registered stop date
So that the persisted result be like: 

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/d03cd6f6-8897-4b2b-8c00-5a14ba0c479d)

IreneDataHub:
Test B-002 22.03.2022/IRS
https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/d03cd6f6-8897-4b2b-8c00-5a14ba0c479d
Create subscription  Stopupdate start 1.11.2022 
Update 1.12. 2022 new description
Stop 1.1 2023 
![image.png](https://images.zenhubusercontent.com/612f6568004bdd03eab7ca46/327c166d-ac35-4493-b1c0-4b67b258ad58)
Update 15. 12.2022 new description "...update 4" 

![image.png](https://images.zenhubusercontent.com/612f6568004bdd03eab7ca46/bc90a023-a3f2-466f-b1ac-f04c49fd2525)
 
Then period two's end date is adjusted
And the master data update is inserted as a third period with the registered stop date= OK

{
                "Id": "6747b8c3-6cdd-47c6-ba18-8428fdd3760d",
                "ChargeId": "89a7787d-f6eb-4184-b3a4-58a802415e3d",
                "TransparentInvoicing": false,
                "Description": "The charge description stop 1 update 4",
                "Name": "demo sub stop and update",
                "VatClassification": 2,
                "StartDateTime": "2022-12-14T23:00:00",
[...]

PR #1214, opened 2022-03-18
1154 update before a registered stop

x-platformcoder:
## Description
Makes sure that a charge updates is valid, even if there is a stop date, as long as the charge start date is earlier than the stop date. The previous charge update will be given a new end date and subsequent charge updates are removed

## References

* #1154


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
```diff
@@ -102,12 +102,19 @@ public void Update(ChargePeriod newChargePeriod)
                 throw new InvalidOperationException("Charge update must not have bound end date.");
             }
 
+            var stopPeriod = _periods.FirstOrDefault(p => p.StartDateTime == p.EndDateTime
```
HenrikSommer, line 105: There should always be 0 or 1 stop period, so I suggest:
```suggestion
            var stopPeriod = _periods.SingleOrDefault(p => p.StartDateTime == p.EndDateTime
```

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
```diff
@@ -102,12 +102,19 @@ public void Update(ChargePeriod newChargePeriod)
...
+                                                          && p.StartDateTime > newChargePeriod.StartDateTime);
```
HenrikSommer, line 106: if `p.StartDateTime == p.EndDateTime` and we are in the update-scenario, I think `p.StartDateTime > newChargePeriod.StartDateTime` is always true? - or that would be inconsistent data, that should result in throw `InvalidOperationException`.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Charges/ChargeTests.cs
```diff
@@ -190,6 +190,91 @@ public void UpdateCharge_WhenEndDateIsBound_ThenThrowException()
             Assert.Throws<InvalidOperationException>(() => sut.Update(newPeriod));
         }
 
+        [Fact]
+        public void UpdateCharge_NewPeriodStartsBeforeExistingStopDate_SetsNewEndDateForExistingPeriodAndInsertsNewPeriod()
+        {
+            // Arrange
+            var stopDate = InstantHelper.GetTodayPlusDaysAtMidnightUtc(4);
+            var existingPeriod = new ChargePeriodBuilder()
```
HenrikSommer, line 198: I suggest we create a private `BuildChargePeriod` method that takes three parameters so that we can remove these details from the test-method to. Build of the variables `existingPeriod`, `stopPeriod` and `newPeriod` would then be one-liners in the test method. The `BuildChargePeriod` method could be reused in the second test-method as well.

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Charges/ChargeTests.cs
```diff
@@ -190,6 +190,91 @@ public void UpdateCharge_WhenEndDateIsBound_ThenThrowException()
...
+            var actualStopPeriod = actualTimeline[2];
+            actualTimeline.Count.Should().Be(3);
+            actualFirstPeriod.Name.Should().Be("ExistingPeriod");
+            actualFirstPeriod.EndDateTime.Should().Be(newPeriodStartDate);
+            actualSecondPeriod.Name.Should().Be("NewPeriod");
+            actualSecondPeriod.StartDateTime.Should().Be(newPeriodStartDate);
+            actualSecondPeriod.EndDateTime.Should().Be(stopDate);
+            actualStopPeriod.StartDateTime.Should().Be(stopDate);
+            actualStopPeriod.EndDateTime.Should().Be(stopDate);
+        }
```
HenrikSommer, line 233: Maybe we should also assert the name of the `actualStopPeriod` is "StopPeriod"?

HenrikSommer (approved):

PR #1222, opened 2022-03-23
Jonasdmoeller/cancel stop new period

jonasdmoeller:
## Description

Changed implementation of cancel stop event. The new period after the stop date now has values from the charge operation instead of extending the existing previous event, as depicted by the diagram in the rule set:

image


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Charges/ChargeTests.cs
```diff
@@ -377,35 +377,57 @@ public void StopCharge_WhenPointsExistAfterStopDate_PointsRemoved()
         public void CancelStop_WhenStopPeriodExists_ThenReplaceEndDateWithDefaultEndDate()
```
prtandrup, line 377: Should the last part of the test name be changed to reflect that the new period is added?

prtandrup (approved):
One comment, but LGTM

HenrikSommer (approved):

Issue #1151, opened 2022-02-10
Persist Charge (1/4) - Update flow


prtandrup:
**Synopsis:**
As a Grid Access Provider,
I want to update my charges with new master data,
So that I may correct mistakes or keep the master data aligned with my own system

This is the first of a few stories with the goal of implementing the [charge persistence rule set](https://github.com/Energinet-DataHub/geh-charges/tree/main/docs/process-flows#persist-charge) (master data only)

With this story we will implement the **update** flow, highlighted below:
<hr>

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/6b9cdd55-a93d-4693-8b37-c06e6c235f65)

<hr>

**Notes:**
Out of scope: Publishing [ChargePeriodUpdated](https://github.com/Energinet-DataHub/geh-charges/blob/main/docs/contracts.md#chargeperiodupdated) integration event

**Acceptance criteria:**

- [ ] 1 Given a charge exist
When processing a charge update request with effective date `after` the charge was created
Then the persisted result must be like this:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/72cf10a3-b8e8-46bb-b17c-f5ef424071e6) <br>

- [ ] 2 Given a charge exist with 3 master data periods
When processing a charge update request with effective date `inside` existing period (1) 
Then the persisted result must be like this:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/6b3dacd8-f01b-4b9b-8803-b621608bf380) <br>

- [ ] 3 Given a charge exist with 3 master data periods
When processing a charge update request with effective date `equal to` period (2) start date
Then the persisted result must be like this:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/81398847-1ff8-4ea5-910c-15b8aa37cb4c) <br>

- [ ] 4 Given a charge exist
When processing a charge update request with effective date `before` the charge was created
Then the persisted result must be like this:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/ff65b35a-b242-465b-9630-f7170e610df4)

- [ ] 5 AC1-4 is done as System Operator with a tax tariff, verify that all grid access providers receive a notification containing the update (Only effectiveDate, no terminationDate in CIM XML)

<hr>

**Task list:**

[...]

Issue #1143, opened 2022-02-07
Ensure a single message with multiple operations for the same charge is processed chronologically (from top to bottom in CIM message) (ac11)


prtandrup:
**Synopsis:** 
As a Grid Access Provider or System Operator
I want to bundle multiple charge operations for the same charge in a single charge request (other charges may included as well)
So that I can minimize the number of requests I need to send

**Tech note:**
* Agreed with the Market: The chronological order is top-down in the request, i.e. the earliest operation must be placed first.
* Currently all charge payloads in a charge request are splitted into separate `ChargeCommand`s and processed. This causes parallel processing, where a charge operation may be misinterpreted as a 'CREATE' operation, when it in fact is an 'UPDATE' operation.

**Out of scope:**
At this point the functionality for stopping a charge may not be implemented, in which case makes stop out of scope.

**Acceptance criteria:**
- [x] (1) Given a charge request with two operations (payloads) for the same charge
When processing the request
Then both operations are persisted showing the chronological order (reading from top to bottom in xml)
<br>

- [x] (2) Given a charge request with three operations (payloads) for the same tax tariff has been persisted
When a Grid Access Provider peeks the notification message
Then it contains all operations, listed in the same chronological order as they were received.
<br>

- [x] (3) Given a charge request with three operations (payloads) for the same charge, where operation1 is **valid**, operation2 is **invalid** and operation3 is **valid**
When DataHub has processed this request
Then operation1 is persisted and a confirmation + notification is available
And operation2 is NOT persisted and a rejection is available 
And operation3 is NOT persisted and a rejection is available (reason being operation2 is rejected)
<br>

- [x] (4) Given a charge request with multiple valid operations for one charge and operations for other charges
When processing the request
Then all operations are persisted
And the following messages sent: confirmation + notification

HenrikSommer:
Todo:
* [X] Add list of operations to ChargeCommand class
* [X] Group identical charge operations in one ChargeCommand - Operations for different charges in the same bundle are split into different ChargeCommands
* [ ] Handle create/update in ChargeCommandReceivedEventHandler when no earlier identical charge is found
* [X] ValidationRules work on an operation
* [X] Rejection flow for an operation (MessageHub)
* [X] Tests must build and pass!

HenrikSommer:
[...]

PR #1189, opened 2022-03-09
Collection of charges in ChargeCommand

HenrikSommer:
## Description

Multiple ordered operations for a charge can be submitted in one CIM/XML. To ensure correct processing of the operations we must handle the operations in the order they were received, which is achieved with the addition of an ordered list of operations on the `ChargeCommand` that is passed on throughout the charges domain.

With this strategy we
* [X] Preserve the order in which operations are handled within the charges domain
* [x] Preserve the order of operations in the confirmation flow
* [x] Preserve the order of operations in the rejection flow (@x-platformcoder)
* [x] Preserve the order of availability notifications 
* [x] Preserve the order in inter-domain communication (@x-platformcoder)
* [x] Inject ChargeOperationDto in unittests (@prtandrup)

## References

* #1151
* #1143

PR #1229, opened 2022-03-24
SPIKE - DRAFT - Maintain charges in parallel using operation intention

HenrikSommer:
## Description

Using the `OperationType`, the `ReceivedDateTime` of the request, and the `ReceivedOrder` of the operations in the CIM/XML bundle we can maintain charges without the need to try and load the charge from the database in order to decide whether the current operation is a `Create`, `Update` or `Stop` operation. The `CancelStop` operation still implies the need to load the charge from the database. This is considered a non-issue.

Tasks:
 * [X] Avoid the necessity of handling operations in the correct sequence in order to save a correct charge timeline, enabling parallel handling of multiple operations on the same charge in a CIM/XML bundle and/or multiple requests
 * [ ] Examine all possible outcomes of `Create`, `Update`, `Stop` and `CancelStop` operations and introduce unit test for each case
 * [ ] Integration test with load to secure validity and reliability of the proposed solution
 * [ ] Preserve operation-order in `AvailableChargeData` table(s) to secure correct ordering of operations in outgoing bundle(s) through MessageHub
 * [ ] Preserve operation-order in integrationevents (maybe we just pass on `OperationType`, `ReceivedDateTime` and `ReceivedOrder` in the integrationevent)

**This spike PR should not be merged to main for the following reasons**
* It presumes the existence of an `OperationType` with the intent of a charge operation in the CIM/XML received from an actor
* It implements its own instance of the `IXmlSchema` of the `Energinet.DataHub.Core.SchemaValidation` NuGet package

## References

* #1151
* #1143


HenrikSommer:
This spike is rejected in favor of #1189

Issue #1217, opened 2022-03-22
Charge - Not possible to stop a charge on creation date - Exception thrown


prtandrup:
**Issue:**
A charge is created with start date May 1. 2022
Then a charge stop request is received, also with May 1. 2022
The system throws an exception (System.InvalidOperationException)

App insights [log](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22SubscriptionId%22%3A%223c13ff3d-aad8-4ee8-94c6-8a938ace9699%22%2C%22ResourceGroup%22%3A%22rg-DataHub-SharedResouces-U-001%22%2C%22Name%22%3A%22appi-shared-sharedres-u-001%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252F3c13ff3d-aad8-4ee8-94c6-8a938ace9699%252FresourceGroups%252Frg-DataHub-SharedResouces-U-001%252Fproviders%252FMicrosoft.Insights%252Fcomponents%252Fappi-shared-sharedres-u-001%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel/%7B%22eventId%22%3A%22fd68476a-a9b8-11ec-997c-2818787ac2f3%22%2C%22timestamp%22%3A%222022-03-22T08%3A18%3A50.633Z%22%2C%22cacheId%22%3A%22b753bc40-c038-48ba-a194-6095e1d5be43%22%2C%22eventTable%22%3A%22requests%22%2C%22timeContext%22%3A%7B%22durationMs%22%3A86400000%2C%22endTime%22%3A%222022-03-22T08%3A34%3A59.721Z%22%7D%7D). 

Problem is rooted here: `Charge.cs`

````
        private void StopExistingPeriod(Instant stopDate)
        {
            var previousPeriod = _periods
                .SingleOrDefault(p =>
                    p.EndDateTime >= stopDate &&
                    p.StartDateTime < stopDate);

            if (previousPeriod == null)
            {
                throw new InvalidOperationException("Cannot stop charge. No period exist on stop date.");
            }
````

**Expected:**
Given a charge is created 1. may 2022
When a stop charge request is received for the same date
Then the stop is processed.
[...]

PR #1233, opened 2022-03-28
Bug fix: Remove chargePeriod when StopDate equals single existing chargePeriod

HenrikSommer:
## Description

It must be possible to stop a charge with only one existing period.
To do this, we have agreed to remove the only existing period.

## References

* #1217


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
```diff
@@ -149,10 +149,10 @@ public void CancelStop(ChargePeriod chargePeriod)
 
         private void StopExistingPeriod(Instant stopDate)
         {
-            var previousPeriod = _periods
-                .SingleOrDefault(p =>
+            var previousPeriod = _periods.OrderByDescending(p => p.Description)
```
jonasdmoeller, line 152: Why order by description?

jonasdmoeller (changes requested):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
HenrikSommer: That's an error! Should have been StartDateTime

jonasdmoeller (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
```diff
@@ -149,10 +149,10 @@ public void CancelStop(ChargePeriod chargePeriod)
 
         private void StopExistingPeriod(Instant stopDate)
         {
-            var previousPeriod = _periods
-                .SingleOrDefault(p =>
+            var previousPeriod = _periods.OrderByDescending(p => p.StartDateTime)
+                .FirstOrDefault(p =>
                     p.EndDateTime >= stopDate &&
-                    p.StartDateTime < stopDate);
+                    p.StartDateTime <= stopDate);
```
prtandrup, line 155: I see a case here, that we do not handle correctly;

Two periods exist:
* Period 1: Ends at day 5
* Period 2: Starts at day 5, Ends at day 9

Stop date received = day 5

as I see it, period 2 will be selected by FirstOrDefault, after which lines 165-167 will create an incorrect newPreviousPeriod row where start and end dates are equal?

prtandrup (changes requested):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
HenrikSommer: I agree, so I will make that change.
We still need `RemoveAllSubsequentPeriods()` for the case, when a stop is added prior to any later period(s).

prtandrup (approved):

PR #1247, opened 2022-04-01
Temporarily introduce an error in a system test to make it fail in Charges CD

prtandrup:
## Description

This PR sounds a bit wierd, but the idea is to test that the system test running after deployment to U-001 will guard further deployments if it fails.

So this PR makes a small correction to a system test, that will make it fail when running in Charges CD.
And will be followed by another PR that rectifies it again.

## References

* #951


prtandrup:
Just got info on a better way to do this, so only one PR is needed. Closing this.

PR #1268, opened 2022-04-06
SPIKE - DONT MERGE - Concurrency handling on atomic charge operations

prtandrup:
## Description

This Spike:
- Use Entity Framework Core's [concurrency handling](https://docs.microsoft.com/en-us/ef/core/saving/concurrency) mechanism to avoid simultaneous update/delete actions to the same entity (read: Charge)
- Adds a unique constraint to the Charges table, to block for multiple threads attempting inserts. 

Currently the spike demonstrates a way to handle above cases in the `ChargeRepositoryTests.cs`, but further down the line this handling should be moved to the `ChargeCommandReceivedEventHandler.cs`
In order to do so, some sort of ´versioning´ is required to ensure the charge time line is placed correctly. Ideas could be

* Add new field to CIM XML, e.g. `RegistrationDateTime`
* Use `RequestDateTime` set by the system, which can be implemented to handle bundles correctly, but simultaneous charge request may have an issue.

## References

* #1143


prtandrup:
Closing this SPIKE PR. Concurrency handling on atomic charge operations is possible, but we have decided not to go with atomic operations (received for the same charge in the same message). In order to process such concurrency issues, we would need to add a new field to the request, like a `registrationDateTime` or such, so the charge timeline can be built as intended by the sender (market participant)

PR #1270, opened 2022-04-06
Update FeeMonthlyPriceSample.xml

IreneDataHub:
## Description

deleted price linies, so  the example match business reason code D18
## References

* #0000


Mech0z (approved):

Mech0z (approved):

Mech0z:
Maybe we should rename the filename to have master data in it? Otherwise one might think this is the price update file

PR #1271, opened 2022-04-06
Update TariffHourlyPricesSample.xml

IreneDataHub:
## Description

deleted price linies, so  the example match business reason code D18


Mech0z (approved):
Maybe we should rename the filename to have master data in it? Otherwise one might think this is the price update file

prtandrup:
Replaced by this [PR](https://github.com/Energinet-DataHub/geh-charges/pull/1586)

PR #1274, opened 2022-04-07
Fix/refac charge command builder

jonasdmoeller:
## Description

Refactored ChargeCommandBuilder() to build only member data, and expanded utilities with separate builder for DocumentDto.

## References

* #0000

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Models/AvailableChargeData/AvailableChargeDataFactoryTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Models/AvailableChargeData/AvailableChargeDataFactoryTests.cs
@@ -97,8 +97,11 @@ namespace GreenEnergyHub.Charges.Tests.MessageHub.Models.AvailableChargeData
         }
 
         [Theory]
-        [InlineAutoDomainData]
+        [InlineAutoDomainData(false, 0)]
+        [InlineAutoDomainData(true, 1)]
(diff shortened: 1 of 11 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Models/AvailableChargeData/AvailableChargeDataFactoryTests.cs
```diff
@@ -100,7 +104,11 @@ public class AvailableChargeDataFactoryTests
             AvailableChargeDataFactory sut)
         {
             // Arrange
-            var chargeCommand = chargeCommandBuilder.WithTaxIndicator(false).Build();
+            var chargeOperationDto = new ChargeOperationDtoBuilder()
+                .WithPoint(1, 1)
+                .WithTaxIndicator(true)
```
prtandrup, line 109: According to the test name it tax should be false. 
Not sure why the actual list is still empty, when it is tax 🤔

prtandrup (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Models/AvailableChargeData/AvailableChargeDataFactoryTests.cs
jonasdmoeller: Thats well spottet. I have updated the test to check both conditions of the tax indicator boolean.

Issue #1276, opened 2022-04-08
Multiple operations for the same charge in a request are not processed correctly


prtandrup:
### Current Behavior
Submitted a charge request (D18) with 3x operations for the same tax tariff (charge)
* Operation 1 (creating the charge)
* Operation 2 (updating)
* Operation 3 (stopping)

Database shows that Operation 1 + 2 are both processed as _creates_
While operation 3 is not processed, an exception is thrown due to 2x charges returned from database, when the stop is to be processed, and only one was expected. This is logged, see link under logs.

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/363ad890-602f-4238-ac08-31e47953643e)

### Expected behavior
Operation 1 treated as a create operation
Operation 2 as an update
Operation 3 as a stop.

### How to reproduce
Use below charge message to reproduce. Remember to change Charge ID (mrid)

````
<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
    <cim:mRID>DocId{{$isoTimestamp}}</cim:mRID>
    <cim:type>D10</cim:type>
    <cim:process.processType>D18</cim:process.processType>
    <cim:businessSector.type>23</cim:businessSector.type>
    <cim:sender_MarketParticipant.mRID codingScheme="A10">{{GridAccessProvider}}</cim:sender_MarketParticipant.mRID>
    <cim:sender_MarketParticipant.marketRole.type>DDM</cim:sender_MarketParticipant.marketRole.type>
    <cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
    <cim:receiver_MarketParticipant.marketRole.type>DDZ</cim:receiver_MarketParticipant.marketRole.type>
    <cim:createdDateTime>{{ISO8601Timestamp}}</cim:createdDateTime>
    <cim:MktActivityRecord>
        <cim:mRID>OpId1{{$isoTimestamp}}</cim:mRID>
        <cim:ChargeGroup>
            <cim:ChargeType>
[...]

PR #1280, opened 2022-04-11
Fix: Handling multiple operations for same charge

prtandrup:
## Description

It was discovered that multiple operations for the same charge, received in a single request, were not processed as expected. See referenced issue.

This PR adds a temporary solution to the problem. It is a short-sighted solution, as the `ChargeCommandReceivedEventHandler.cs` is being refactored (to decrease its' responsibility and make the code cleaner). 

Although not preferred, this solution violates the unit-of-work pattern by saving the `create` charge operation, in order for the next operations to be business validated correctly (read: otherwise `ChargeCommandBusinessValidationRulesFactory.cs` would not know of an existing charge)

With this PR, the ChargeCommandReceivedEventHandler 

* Processes the first operation and in case it is a `create` operation, it's stores it and an accepted domain event is sent before moving on to the next operation (for the same charge).
* The next operations are processed together (one by one) before persisting and sending out accepted domain events (or rejected events).

## References

* #1276

### DIFF
--- a/.github/actions/dotnet-prepare-outputs/action.yml
+++ b/.github/actions/dotnet-prepare-outputs/action.yml
@@ -1,0 +1,73 @@
+# Copyright 2020 Energinet DataHub A/S
+#
+# Licensed under the Apache License, Version 2.0 (the "License2");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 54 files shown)


x-platformcoder (approved):
LGTM

HenrikSommer (approved):

Issue #1316, opened 2022-04-29
Synchronous error response must match the agreed format, i.e. use xml tags and codes


prtandrup:
As a Market Participant submitting invalid requests* to the charges domain
I want to receive synchronous error that complies to the agreed format
So that my system may process it correctly.

\* Violating:
* Schema validation
* Sender provided in body does not match organization in bearer token

Agreed format for a synchronous error response can be found [here](https://energinet.dk/-/media/982EDBF70CBC4623B813A7101A14C722.pdf), see section "_2.3 Validering af forespørgsler_" 

Currently our response is not wrapped in xml tags as seen here:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/f5ba09d8-b78f-4736-9904-6bff8e8a465a)

**Notes:**
Please investigate if we are to use the existing B2B-codes used by DataHub2 or not.

**Acceptance criteria:**

- [x] 1 Given a request is made that **violates schema validation** (both for Charges and Charge Links)
When receiving a synchronous **bad request** response
Then the error body is formatted according to the **agreed format**
And using the correct **B2B-005** code
<br>

- [x] 2 Given a request is made that **violates 'Sender mismatch'** (both for Charges and Charge Links)
When receiving a synchronous **bad request** response
Then the error body is formatted according to the **agreed format** 
And uses the correct **B2B-008** code

prtandrup:
### Tested in T-001:

**When violating schema - Charges & Charge Links - Reponse:**

````
<?xml version="1.0" encoding="utf-8"?>
<Error>
  <Code>BadArgument</Code>
  <Message>The specified input does not pass schema validation.</Message>
  <Details>
    <Error>
      <Code>SchemaValidationError</Code>
      <Message>The element 'RequestChangeBillingMasterData_MarketDocument' in namespace 'urn:ediel.org:structure:requestchangebillingmasterdata:0:1' has invalid child element 'MktActivityRecord' in namespace 'urn:ediel.org:structure:requestchangebillingmasterdata:0:1'. List of possible elements expected: 'createdDateTime' in namespace 'urn:ediel.org:structure:requestchangebillingmasterdata:0:1'.</Message>
      <InnerError>
        <LineNumber>13</LineNumber>
        <LinePosition>6</LinePosition>
      </InnerError>
    </Error>
  </Details>
</Error>
````

:warning: `BadArgument` is used instead of `B2B-005`, when violating schema. 
[...]

PR #1336, opened 2022-05-03
1316 synchronous error response must match the agreed format

x-platformcoder:
## Description

Synchronous error response must match the agreed format, i.e. use xml tags and codes#1316
As a Market Participant submitting invalid requests* to the charges domain
I want to receive synchronous error that complies to the agreed format
So that my system may process it correctly.

* Violating:

Schema validation
Sender provided in body does not match organization in bearer token
Agreed format for a synchronous error response can be found [here](https://energinet.dk/-/media/982EDBF70CBC4623B813A7101A14C722.pdf), see section "2.3 Validering af forespørgsler"

Currently our response is not wrapped in xml tags as seen here:

image.png

Notes:
Please investigate if we are to use the existing B2B-codes used by DataHub2 or not.

Acceptance criteria:

[ ] 1 Given a request is made that violates schema validation (both for Charges and Charge Links)
When receiving a synchronous bad request response
Then the error body is formatted according to the agreed format
And using the correct B2B-code

[ ] 1 Given a request is made that violates 'Sender mismatch' (both for Charges and Charge Links)
When receiving a synchronous bad request response
Then the error body is formatted according to the agreed format
And uses the correct B2B-code
## References

* #1316


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure.Core/Function/HttpResponseBuilder.cs
```diff
@@ -46,13 +48,28 @@ public HttpResponseData CreateAcceptedResponse(HttpRequestData request)
             return httpResponse;
         }
 
-        public HttpResponseData CreateBadRequestWithErrorText(HttpRequestData request, string errorText)
+        public HttpResponseData CreateBadRequestResponseWithText(HttpRequestData request, string errorText)
         {
```
HenrikSommer, line 52: We should add tests covering this method and the private `CreateErrorMessageAsXml()` method.

x-platformcoder: done

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure.Core/Function/HttpResponseBuilderTests.cs
```diff
@@ -77,6 +78,32 @@ public void CreateAcceptedResponse_Creates_AcceptedResponseWithCorrelationId(Fun
             responseData.StatusCode.Should().Be(HttpStatusCode.BadRequest);
         }
 
+        [Theory]
+        [InlineAutoMoqData]
+        public void CreateBadRequestResponseWithText_Creates_Response_With_Text_In_Xml_Body(
```
HenrikSommer, line 83: Elsewhere in the codebase we seek to name test-methods using the Gherkin syntax (Given_When_Then) and i suggest: 
```suggestion
        public void CreateBadRequestResponseWithText_WhenBadRequest_CreatesResponseWithTextInXmlBody(
```

HenrikSommer (approved):
Approved with a suggestion

HenrikSommer (approved):

Issue #1340, opened 2022-05-05
MessageType HTTP headers are incorrect for confirmations and rejections (charge + links)


prtandrup:
The HTTP header MessageType does not contain correct values for Charge + Charge Links confirmations and rejections.
According to SME Christian, the values shall match the messages' CIM names, located in this [guide](https://energinet.dk/-/media/71818148C5ED436F87A032D5DDE06F1F.pdf).

**Acceptance criteria**

Given an actor peeks one of below messages,
When checking the HTTP header 'MessageType'
Then the value must be as listed below:

|Message|MessageType
|---|---|
|Charge confirmation|ConfirmRequestChangeOfPriceList|
|Charge rejection|RejectRequestChangeOfPriceList|
|Charge link confirmation|ConfirmRequestChangeBillingMasterData|
|Charge link rejection|RejectRequestChangeBillingMasterData|

prtandrup:
Deployed PR 1343 as a prerelease to U-001 and verify that it didn't conflict with existing data in MessageHub receipt table. As expected, existing data will use the incorrect header value, but data received after deployment use the new correct value. 

**Confirm:** 
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/8f8d9af2-23f9-481d-b326-79d3164947bb)

**Rejection:**
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/c9e321a7-7151-4a54-9a44-7b2b751080f2)

prtandrup:
Charge Link part deployed as well, here seen on U-001:

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/34af3d93-4332-4187-8b04-3d8fba9889b6)

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/2470ecda-bbba-4908-a386-01cf4f19df80)

prtandrup:
**For the tester:**

The Charges part can easily be tested on T-001.
Charge Links however, might be best doing in U-001 og B-002.

IreneDataHub:
Test 10 May: Use environment  t-001
Acc. 
Given an actor peeks one of below messages:
Message	MessageType
Charge confirmation	ConfirmRequestChangeOfPriceList
Charge rejection	RejectRequestChangeOfPriceList
Charge link confirmation	ConfirmRequestChangeBillingMasterData
Charge link rejection	RejectRequestChangeBillingMasterData

[...]

PR #1343, opened 2022-05-05
Fix: Use correct `MessageType` HTTP header values for Charge `confirmations` and `rejections`

prtandrup:
## Description

At the moment, when a market partipant peeks a charge confirmation or rejection message an incorrect value is set in the `MessageType` HTTP header.

With this PR, the MessageType for Charge Confirmation will be set to `ConfirmRequestChangeOfPriceList` whereas Rejection will be `RejectRequestChangeOfPriceList`

## References

* #1340


x-platformcoder (approved):
LGTM

Issue #1339, opened 2022-05-04
Validations in charge & chargelinks ingestions are throwing misleading error messages or not creating any reject notification regarding to documentvalidation errors


x-platformcoder:
As a Market Participant submitting a charge or chargelink document to the charge domain that should result in a schemavalidation error or authorization error.

I Want an explicit error message that tells me exactly what went wrong instead of the misleading error that is passed now.

So that i can take action either by submitting new documents or raising an incident.

**Current behaviour**
Given we submitting a RequestChangeOfPriceList_MarketDocument with an invalid business reason code we receive the synchronous errormessage "The sender organization provided in the request body does not match the organization in the bearer token.". (see picture)

Given we submitting a valid MarketDocument but the marketparticipant actor in the chargesdomain do not exist, we receive the synchronous errormessage "The sender organization provided in the request body does not match the organization in the bearer token.".

Given we submitting a valid MarketDocument and the business reason code is known, but is invalid in this context eg. D17, we do not receive a rejection notification.

**Notes**

Current documentvalidations:

* CommandSenderMustBeAnExistingMarketParticipantRule      -> scenario already fails in ingestionfunction (400 Bad Request), how is that possible?
* BusinessReasonCodeMustBeUpdateChargeInformationRule     -> unknown codes fails in schemavalidation (400 Bad Request)
* DocumentTypeMustBeRequestUpdateChargeInformationRule    -> scenario already fails in schemavalidation (400 Bad Request)
* RecipientIsMandatoryTypeValidationRule                  -> scenario already fails in schemavalidation (500 Server error)
* SenderIsMandatoryTypeValidationRule                     -> scenario already fails in schemavalidation (500 Server error)   

None of the above is triggered by the DocumentValidator, all has failed in the ingestionfunction

* BusinessReasonCodeMustBeUpdateChargeInformationRule     -> known codes but invalid in current document, does not trigger a rejection notification

This one should trigger a rejection notification by the DocumentValidator, but it does not happend

This bug is related to story #1230 & #1305

**Acceptance criteria:**
- [ ] Schemavalidations should return the correct schemavalidation error instead
- [ ] Should return an explicit errormessage if received actorcontext is invalid
- [ ] Validations in this order -> SchemaValidation, AuthenticatedSenderIdMatchesDocumentSenderId
- [ ] Same criterias for both charge and chargelinks ingestions

**Input and Result**
[...]

PR #1347, opened 2022-05-09
fix: document validation for update charge prices

Mech0z:
## Description

Nullchecker and a businessrule currently blocks flow of update charge prices

## References

* #1339


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Dtos/ChargeCommands/ChargeCommandNullCheckerTests.cs
```diff
@@ -45,6 +46,7 @@ public class ChargeCommandNullCheckerTests
                 .Build();
             var documentDto = new DocumentDtoBuilder()
                 .WithDocumentId(documentId)
+                .WithBusinessReasonCode(BusinessReasonCode.UpdateChargeInformation)
```
HenrikSommer, line 49: Should we add a test case with `BusinessReasonCode.ChargePrices` or is that case covered?

PR #1350, opened 2022-05-10
Log validation errors from DataAvailableNotifiers

HenrikSommer:
## Description

As a developer, it is of high value to receive logging information regarding valid errors (non-exceptions). This pull request introduces logging of asynchronous validation errors, which is of high value, especially during troubleshooting of failed integration tests but also for troubleshooting failed requests to a live environment.


prtandrup (approved):
LGTM

Issue #1291, opened 2022-04-22
Charge price lists: Be able to reject an invalid request (Feature-AC4b/4c)


prtandrup:
As a Charge Owner
I want to receive an error message for an invalid charge price list containing an error code and error text 
So that I may correct and try again

**Note:**
It is assumed that first validation done on Charge Prices requests (business reason code D08) will be **document validation**, which is being implemented by story #1305.

**Acceptance criteria:**
- [ ]  1 Given a market participant submits a Charge Price request that contains **an error in the document header**
When DataHub has processed it
Then the sender may peek a **rejection** message which...
* uses the business reason code **D08**
* contains **an error code**
* contains **error text**
<br>

- [ ] 2 Given a market participant submits a valid Charge Price request 
When DataHub has processed it
Then the sender receives a **confirmation** message.
<br>

- [ ] ~~3 Architecture diagram for Charge Prices has been updated to reflect any changes that this story introduced.~~

prtandrup:
CommandSenderMustBeAnExistingMarketParticipantRule

PR #1353, opened 2022-05-10
feat: validation of recipient

Mech0z:
## Description

Checks that recipient role is always DDZ

## References

* #1291


x-platformcoder (approved):
LGTM

Issue #1345, opened 2022-05-06
Switch to notify MessageHub with ActorId instead of GLN number when notifying about new messages


prtandrup:
**Synopsis**
Instead of GLN which is not unique anymore we are switching to use internal Actor ID

As a Charges domain owner
I want the MessageHub integration to use the newest agreed standards
So that the Charges domain uses the latest version.

**Tech Notes**
The Actor ID can be fetched from token via GetClaim(claimsPrincipal.Claims, "azp"); see ActorMiddleware or ActorProvider for inspiration
Currently GLN is saved in [MessageHub].[AvailableChargeData] as RecipientId and their role as RecipientRole, these 2 can via a lookup be changed to the internal actor id [Charges].[MarketParticipant].[Id]

**Acceptance criteria:** 

- [x] When interacting with the MessageHub domain we use ActorID instead of GLN

Titans dependency:
[Story 548](https://app.zenhub.com/workspaces/titans-60d029726fc7570010362aa7/issues/energinet-datahub/geh-post-office/548)

New package https://www.nuget.org/packages/Energinet.DataHub.MessageHub.Client/3.0.0

HenrikSommer:
Test ok!
It is still possible to peek messages through MessageHub following the shift to provide ActorId instead of GLN to MessageHub in the _dataavailable_ message.
The code that does the magic is found in the _CreateNotifications_ method in the _AvailableDataNotifier_ class.

HenrikSommer:
Seems like there's a bug in MessageHub. Titans are working on it, but we are blocked until further notice.

prtandrup:
## Tests are carried out in U-001, T-001, B-002

### Test results for U-001

* Charge confirmation: :white_check_mark:
* Charge notification: :white_check_mark:
* Charge rejection: :white_check_mark:

* Charge Link confirmation: :white_check_mark:
* Charge Link notification: :white_check_mark:
* Charge Link rejection: :white_check_mark:

T and B will follow.

PR #1363, opened 2022-05-18
feat: 1345 Switch to notify MessageHub with ActorId instead of GLN number when notifying about new messages

x-platformcoder:
## Description
Availabledata has been extended to hold an actor id, as it is necessary to notify the messagehub with an actor id instead of only the GLN number from messagehub.client version 3.0.0 and up

## References

* #1345


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.MessageHub/MessageHub/AvailableDataNotificationFactory.cs
```diff
@@ -30,12 +30,12 @@ public class AvailableDataNotificationFactory<TAvailableData> : IAvailableDataNo
             return availableData.Select(
                     data => new DataAvailableNotificationDto(
                         data.AvailableDataReferenceId,
-                        new GlobalLocationNumberDto(data.RecipientId),
+                        new ActorIdDto(data.ActorId),
```
HenrikSommer, line 33: Lines #33, 35 and 38 are not covered by tests.

HenrikSommer (commented):
Let's add a few tests `AvailableDataNotificationFactory` and we're good to go

HenrikSommer (approved):

PR #1373, opened 2022-05-30
Key vault reference

dstenroejl:
## Description

Fix bug where secrets located in Key Vault is used in clear text.


prtandrup (approved):
LGTM

PR #1374, opened 2022-05-30
LRN/DBH Break stuff

lasrinnil:
## References

* #0000


lasrinnil:
@prtandrup check this it. Would this be ok?

prtandrup:
Thanks for the contribution. I've had a look at the problem raised. Turns out adding `=` in our business logic does not have any effect at the outcome, as the subsequent method will handle it anyways. However, if you change `<` to `>`, then some of the tests will fail. 
image

PR closed.

PR #1376, opened 2022-05-30
Change OrderBy

mbusk88:
## References

* #0000


prtandrup:
Thanks for the contribution. 
After looking into it, it appears that `OrderByDescending` has no relevancy. The outcome is still the same when it is `OrderBy` (thanks to this other private method `RemoveAllSubsequentPeriods`).

PR [1440](https://github.com/Energinet-DataHub/geh-charges/pull/1440) removes the `OrderByDescending`,

PR #1380, opened 2022-05-30
LRN/DPH Breakstuff 2

lasrinnil:
## References

* #0000

### DIFF
--- a/source/GreenEnergyHub.Charges/global.json
+++ b/source/GreenEnergyHub.Charges/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.201",
+    "version": "6.0.300",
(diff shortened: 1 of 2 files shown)


lasrinnil:
@Mech0z Would this be ok?

prtandrup:
Thanks for the contribution 👍 
Based on it, global.json for both solutions has been bumped to newest SDK 6.0.300 (already in main)
And the unit test that should have caught below change in `Charge.cs`, has been refactored in another [PR](https://github.com/Energinet-DataHub/geh-charges/pull/1440). 
The test `StopCharge_WhenPointsExistOnAndAfterStopDate_PointsRemoved()` will now fail if below is done by mistake.
image
PR closed.

Issue #1292, opened 2022-04-22
Persist charge price lists (Feature-AC1b)


prtandrup:
As DataHub
I want to persist the prices received for charges
So the current prices can be provided on request may it be frontend or market participant request (to be implemented)

**Notes:**
* The price lists are to be stored in the ChargePoint table. Should we consider renaming this to ChargePrices?
~~* Today we have a Position column. This should be removed with this story. There is no need to store Position.
Be aware that removing this will impact the Charge Master Data flow, which currently still stores prices as well.~~ moved to #1411

**Acceptance criteria:**

- [x] 1. Persist price list rule set has been defined and published in GitHub Repo, [here](https://github.com/Energinet-DataHub/geh-charges/tree/main/docs/process-flows#persist-charge-prices)
<br>

- [x] 2 Given a charge owner has registered a charge in DataHub
When a price list is submitted for this charge
Then it is persisted in the database
<br>

- [x] 3 Given a charge owner has registered a charge in DataHub
And a price series is already stored
When a price list is submitted covering the same period
Then the stored price lists is updated with the new values (historical values are currently not stored)
<br>

- [x] 4 Given a charge owner has registered a charge in DataHub
And a price series is already stored
When a price list is submitted that covers **a part of** the same period
Then the overlapping prices are updated 
And the non-overlapping prices remain
And the new prices are stored

~~- [ ] 5 Architecture diagram for Charge Prices has been updated to reflect any changes that this story introduced.~~

prtandrup:
### Tested in T-001
:warning: The use of `Position` is still to be removed.

**AC2:** :white_check_mark:

````
Given a charge owner has registered a charge in DataHub
When a price list is submitted for this charge
Then it is persisted in the database
````

It was tested with information like `priceTimeFrame_Period.resolution`, `VATPayer` and `taxIndicator` being present

**Submitted D08 price request:** 
````
<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
	<cim:mRID>DocId2022-06-10T06:08:02.372Z</cim:mRID>
[...]

PR #1387, opened 2022-06-01
fix: Read points timeinterval from CIM document correctly

x-platformcoder:
## Description

Timeinterval from cim document was not read correctly, so startdate for a point timeinterval was inherited from effective date and enddate for a point timeinterval was null, and that resulted in existing points were not overwritten but added as a new set of points.

## References

* #1292


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/CimDeserialization/ChargeBundle/ChargeCommandConverter.cs
```diff
@@ -233,10 +235,11 @@ private async Task<ChargeOperationDto> ParseChargeTypeElementIntoOperationAsync(
                 points);
         }
 
-        private async Task<(List<Point> Points, Resolution Resolution)> ParseSeriesPeriodIntoOperationAsync(SchemaValidatingReader reader, Instant startDateTime, Resolution initialResolution)
+        private async Task<(List<Point> Points, Resolution Resolution, Instant IntervalStartTime, Instant IntervalEndTime)> ParseSeriesPeriodIntoOperationAsync(SchemaValidatingReader reader, Instant startDateTime, Resolution initialResolution)
```
HenrikSommer, line 238: Quite a heavy tuple - maybe it's time for a result class?

PR #1406, opened 2022-06-07
refactor: Move shared elements to SharedDtos

HenrikSommer:
## Description

Some elements were inaccurately placed in the `MarketParticipant` namespace of the charges domain, so we move them to the `SharedDtos` namespace


x-platformcoder (approved):
fine

Issue #1409, opened 2022-06-08
New Price Flow: #1 Introduce separate command for charge price


HenrikSommer:
First story of the new price flow
Split ChargeIngestion so it published both the AS-IS event and the new D08 event.

Given a D08 price request is received
When processed (by ChargeIngestion)
Then the existing ChargeCommandReceivedEvent is published (as is)
And a new D08 event is published on a new service bus D08 topic.

Tasks: 
- [x] Split xml into price and information event (D18/D08)
- [x] New D08 topic on service bus
- [x] Add new function (ChargePriceReceiverEndpoint)
- [x] Add health check

prtandrup:
### Tested in T-001

````
AC 1
Given a D08 price request is received
When processed (by ChargeIngestion)
Then the existing ChargeCommandReceivedEvent is published (as is)
And a new D08 event is published on a new service bus D08 topic.
````

Submitted D08 price series request for subscription id: SubId014:

````
<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
    <cim:mRID>DocId2022-08-02T07:21:04.679Z</cim:mRID>
    <cim:type>D10</cim:type>
    <cim:process.processType>D08</cim:process.processType>
    <cim:businessSector.type>23</cim:businessSector.type>
    <cim:sender_MarketParticipant.mRID codingScheme="A10">8100000000016</cim:sender_MarketParticipant.mRID>
    <cim:sender_MarketParticipant.marketRole.type>DDM</cim:sender_MarketParticipant.marketRole.type>
    <cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
    <cim:receiver_MarketParticipant.marketRole.type>DDZ</cim:receiver_MarketParticipant.marketRole.type>
    <cim:createdDateTime>2022-08-02T07:21:04Z</cim:createdDateTime>
    <cim:MktActivityRecord>
        <cim:mRID>OpId2022-08-02T07:21:04.679Z</cim:mRID>
        <cim:ChargeGroup>
[...]

PR #1432, opened 2022-06-16
docs: Add new `Charge Prices` architecture "TO-BE" diagram

prtandrup:
## Description

This PR adds an architecture diagram for handling Charge Prices, which we are in progress of separating from the Charge Information handling.

Also, minor name changes were made to other architecture diagrams, basically renaming API Gateway to API Management & Aggregation Domain to Wholesale Domain. 

## References

* #1409


HenrikSommer (commented):
Aggregations domain is now Wholesale domain

Issue #1208, opened 2022-03-17
Update market participant data through the use of integration events from the Market participant domain


prtandrup:
As the Charges domain
I want to be able to update the market participant data 
through the use of integration events published by the Market participant domain
So that this becomes an automatic process

**Dependencies:**
- Titans will provide a nuget package
- 1x Titans developer will be available for pair programming for 2 days, helping Volt implementing this and to get feedback.

**Tech notes:**
~~It is assumed that we can continue with our current market participant data in charges and apply integration events on top of that. (this note is obsolete)~~
- Please decide how Volt are to handle fault situations/states

**Acceptance criteria:**
- [x] 1 Charges market participants are automatically updated when integration events are published by the Market participant domain. 
- [x] 2 Remove functionality for manually synchronizing data from Market participant database to Charges database.
- [x] 3 Remove Postman request for synchronizing data
- [x] 4 Throws an exception if market participant with same marketparticipantid is already persisted with another role than the one received in the actor integrationevent
- [ ] ~~5 Throws an exception if the actor integrationevent has multiple roles associated~~

**Task list:**
- [x] Update architecture diagram (charge and charge links) to reflect the Market Participant domain, our new function for consuming the integration events published.

x-platformcoder:
Architecture![CF52F095-8F2D-4527-A2C6-244824DE621A.jpeg](https://images.zenhubusercontent.com/621e27789626c44a088a5ee9/064332cc-bd27-4579-b5a6-580ed4651aa3)![35C8758F-0975-4E92-A31C-7D85F9CEEE1C.jpeg](https://images.zenhubusercontent.com/621e27789626c44a088a5ee9/4834c7ef-f48f-4b40-ae57-6004eaab52c6)

x-platformcoder:
Test af Integration: https://app-webapi-markpart-u-001.azurewebsites.net/swagger/index.html
Der er ikke udstillet noget til GridArea endnu, de har ikke noget der pt. kan oprette disse, det bliver leveret sammen med GridLinks iflg. Jesper fra Titans.

prtandrup:
**Blocked**.
[...]

PR #1434, opened 2022-06-20
fix: throws exception if persisting a marketparticipant with an exisiting id but different role

x-platformcoder:
## Description

If an actor integrationevent from the market participant domain  is received with more than 1 role associated an invalidoperation exception is thrown and no operations is done, the same will happen if a market participant is already persisted with the marketparticipant id that is included in the actor event, but with a different role than the one included in the actor event

## References

* #1208

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MarketParticipants/Handlers/MarketParticipantDomainEventMapper.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MarketParticipants/Handlers/MarketParticipantDomainEventMapper.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System;
 using System.Linq;
 using Energinet.DataHub.MarketParticipant.Integration.Model.Dtos;
 using GreenEnergyHub.Charges.Domain.Dtos.GridAreas;
@@ -28,6 +29,13 @@ namespace GreenEnergyHub.Charges.Application.MarketParticipants.Handlers
 
             var rolesUsedInChargesDomain = actorUpdatedIntegrationEvent.BusinessRoles
                 .Select(MarketParticipantRoleMapper.Map).ToList();
+
+            if (rolesUsedInChargesDomain.Count > 1)
+            {
+                throw new InvalidOperationException(
+                    $"Only 1 role per market participant with ID '{actorUpdatedIntegrationEvent.Gln}' is allowed, " +
+                    $"the current market participant has {rolesUsedInChargesDomain.Count} roles associated");
+            }
 
             return new MarketParticipantUpdatedEvent(
                 actorUpdatedIntegrationEvent.ActorId,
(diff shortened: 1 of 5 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MarketParticipants/Handlers/MarketParticipantDomainEventMapper.cs
```diff
@@ -29,6 +30,13 @@ public static class MarketParticipantDomainEventMapper
             var rolesUsedInChargesDomain = actorUpdatedIntegrationEvent.BusinessRoles
                 .Select(MarketParticipantRoleMapper.Map).ToList();
 
+            if (rolesUsedInChargesDomain.Count > 1)
+            {
+                throw new InvalidOperationException(
+                    $"Only 1 role per market participant with ID '{actorUpdatedIntegrationEvent.Gln}' is allowed, " +
+                    $"the current market participant has {rolesUsedInChargesDomain.Count} roles associated");
```
prtandrup, line 37: nit: perhaps this will come in handy when coming across this error

```suggestion
                    $"Only 1 role per market participant with ID '{actorUpdatedIntegrationEvent.Gln}' is allowed, " +
                    $"the current market participant has {rolesUsedInChargesDomain.Count} roles associated in the integration event with id '{actorUpdatedIntegrationEvent.Id}'");
```

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/MarketParticipants/Handlers/MarketParticipantPersister.cs
```diff
@@ -53,6 +53,16 @@ public async Task PersistAsync(MarketParticipantUpdatedEvent marketParticipantUp
 
                 if (persistMarketParticipant is null)
                 {
+                    var marketParticipantAlreadyExist = await _marketParticipantRepository
```
prtandrup, line 56: nit:

```suggestion
                    var marketParticipantAlreadyExistWithOtherRole = await _marketParticipantRepository
```

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Application/MarketParticipants/Handlers/MarketParticipantPersisterTests.cs
```diff
@@ -64,11 +64,7 @@ public class MarketParticipantPersisterTests
             await sut.PersistAsync(marketParticipantUpdatedEvent).ConfigureAwait(false);
 
             // Assert
-            marketParticipantRepository.Verify(v => v.AddAsync(It.IsAny<MarketParticipant>()), Times.Exactly(2));
-            logger.VerifyLoggerWasCalled(
```
prtandrup, line 68: Why did we remove the log entry? Seen a couple of places

prtandrup (approved):
LGTM, with a few minor comments

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Application/MarketParticipants/Handlers/MarketParticipantPersisterTests.cs
x-platformcoder: Because it is no longer possible to have 2 logentries with 2 different roles in an actor integrationevent eg. EnergySupplier and GridAccessProvider... it will throw an exception (another test)

Issue #1326, opened 2022-05-02
Ensure all required master data fields are present and populated for Charge Information Request (D18) (Feature-AC4d)


prtandrup:
Notes:
Input validation on the following fields. Be aware that some of them might be caught by the schema validation, and thus should probably not be covered in the input validation as well... 

Null or empty checks of all the fields that are allowed set as 0:1 in schema, but are required for Charge Information requests:

- [x] - Name https://github.com/Energinet-DataHub/geh-charges/pull/1437
- [x] - Description https://github.com/Energinet-DataHub/geh-charges/pull/1437
- [x] - Resolution https://github.com/Energinet-DataHub/geh-charges/pull/1439

Content validation

- [x] * Owner => Length 13-16 https://github.com/Energinet-DataHub/geh-charges/pull/1444
- [x] * Type => Valid Enum check https://github.com/Energinet-DataHub/geh-charges/blob/main/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeTypeIsKnownValidationRule.cs
- [x] * Charge Id => Length Max 10 https://github.com/Energinet-DataHub/geh-charges/blob/main/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeIdLengthValidationRule.cs
- [x] * Name => Length limit 132 https://github.com/Energinet-DataHub/geh-charges/pull/1438
- [x] * Description => Max 2048 Length VR.447 https://github.com/Energinet-DataHub/geh-charges/blob/main/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/ChargeDescriptionHasMaximumLengthRule.cs
- [x] * VAT Classification => value provided => not unknown at this point https://github.com/Energinet-DataHub/geh-charges/blob/main/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/ChargeCommands/Validation/InputValidation/ValidationRules/VatClassificationValidationRule.cs
- [x] * Transparent Invoicing => value provided => not unknown at this point https://github.com/Energinet-DataHub/geh-charges/pull/1445
[...]

PR #1438, opened 2022-06-21
fix: Max Charge Name input validation

Mech0z:
Was set to 50, but according to Cim guide the correct is 132

Database already have the correct 132 max length

See https://energinet.dk/-/media/71818148C5ED436F87A032D5DDE06F1F.pdf page 222

## References

* #1326


x-platformcoder (approved):
ok

PR #1440, opened 2022-06-21
Clean-up: `ChargeTests.cs`

prtandrup:
## Description

Refactored / Cleaned-up `ChargeTests.cs` after ChargeBuilder was refactored to only allow Charge creation through the domain model `Charge.cs`

Also, some PRs were opened at our last bug-hunt, which proved that some existing tests didn't catch logic changes in our Charge domain class. The relevant tests have been altered to catch such changes.

Also, in `Charge.cs`, the use of `OrderByDescending` was removed in the private `StopExistingPeriod` method, as it had no real impact on the outcome.

## References

[PR1380](https://github.com/Energinet-DataHub/geh-charges/pull/1380)
[PR1376](https://github.com/Energinet-DataHub/geh-charges/pull/1376)

* #0000


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Domain/Charges/ChargeTests.cs
```diff
@@ -355,9 +345,10 @@ public void StopCharge_WhenPointsExistAfterStopDate_PointsRemoved()
                 new(0, decimal.One, InstantHelper.GetTodayPlusDaysAtMidnightUtc(2)),
                 new(0, decimal.One, InstantHelper.GetTodayPlusDaysAtMidnightUtc(3)),
                 new(0, decimal.One, InstantHelper.GetTodayPlusDaysAtMidnightUtc(4)),
+                new(0, decimal.One, InstantHelper.GetTodayPlusDaysAtMidnightUtc(5)),
```
HenrikSommer, line 348: Why do we need an extra point to end up with `sut.Points.Count.Should().Be(2);`?

PR #1448, opened 2022-06-24
feat: add termination date validation rule

Mech0z:
## Description

When termination date (EndDate) is set it must be the same
as effective date (StartDate)

## References

* #1326


x-platformcoder (approved):
lgtm

PR #1479, opened 2022-07-12
Conditionally check for reject-message

HenrikSommer:
## Description

For now, ChargeLinkCommandConverter splits all CIM MktActivityRecord into separate ChargeLinkCommands
so we do not always receive a rejection due to the parallel handling of commands. This behavior caused a test to be unreliable as the system sometimes returned two confirmed messages and other times one confirmed and one rejected message. Therefore the check for a rejected message is no longer mandatory, but if a rejected message is present then it must be rejected due to a violation of the `ChargeLinksUpdateNotYetSupportedRule`.

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/DomainTests/ChargeLinkIngestionTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/DomainTests/ChargeLinkIngestionTests.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.IntegrationTests/DomainTests/ChargeLinkIngestionTests.cs
```diff
@@ -128,7 +129,11 @@ public async Task Given_NewTaxChargeLinkMessage_When_GridAccessProviderPeeks_The
                 var peekResults = await Fixture.MessageHubMock.AssertPeekReceivesRepliesAsync(correlationId, 3);
                 peekResults.Should().ContainMatch("*ConfirmRequestChangeBillingMasterData_MarketDocument*");
                 peekResults.Should().ContainMatch("*NotifyBillingMasterData_MarketDocument*");
-                peekResults.Should().ContainMatch("*RejectRequestChangeBillingMasterData_MarketDocument*");
+
+                // For now ChargeLinkCommandConverter splits all CIM MktActivityRecord into separate ChargeLinkCommands
+                // so we do not always receive a rejection due to the parallel handling of commands.
```
HenrikSommer, line 134:
```suggestion
                // For now, ChargeLinkCommandConverter splits all CIM MktActivityRecord into separate ChargeLinkCommands
                // so we do not always receive a rejection due to the parallel handling of commands.
```

prtandrup (approved):

PR #1487, opened 2022-07-14
fix: Revert to dotnet-solution-ci 7.5.0 (WIP)

HenrikSommer:
## Description

To overcome an issue running integration tests using Azure resources, we revert to an earlier version of the dotnet-solution-ci without oidc. This is a temporary solution, and should be followed up by a new version of dotnet-solution-ci.


HenrikSommer:
That fix didn't work. Closing PR.

Issue #1411, opened 2022-06-08
New Price Flow: #3 confirm/reject/notify and persist


HenrikSommer:
In the previous story #1410 the orchestration was established for processing the `PriceCommandReceived` event.

With this story, this new price architecture will be responsible for **persisting prices, sending confirmations/rejections** and **notifications**. 
Basically, we switch `ON` handling Charge Price request (D08) through the new Price architecture and at the same time switch `OFF` the charge price handling in the "old" architecture, which is expected to only handle Charge Information (D18) afterwards.

![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/1072374e-d092-4681-afc9-064ba62cc343)

**Tech notes:**
* No downtime allowed for processing D08 charge prices; switching from old to new architecture must be carried out at the same time.
* Outbox pattern is out of scope for this story. 
* Notifying MessageHub about available data must be done without outbox.
* The Integration Event ChargePricesUpdated is out of scope (handled in a subsequent story, and no registered consumers exist, yet)

**Acceptance criteria:** 
- [ ] 1 Given the **new architecture is switched ON** and **the old architecture OFF**
When a Charge Price request (D08) is received
Then it is **processed only by the new architecture** 
(testable through app insights)
<br>

- [ ] 2 Given a **valid** charge price request is received
When the sender peeks
Then a **D08 confirmation** message is provided
<br>

- [ ] 3 Given a valid charge price request for a tax tariff is received
When a Grid Access Provider peeks
Then a **D08 notification** message is provided
<br>

- [ ] 4 Given an invalid charge price request is received
When the sender peeks
Then a **D08 rejection** message is provided

**Tasks**
- [ ] Move confirmation functionality from "old arch" to new service + enable it to provide MessageHub with a "DataAvailableNotification"
- [ ] Same for rejection functionality
- [ ]  Same for notification functionality
- [ ] Be able to persist charge prices from the new architecture
- [ ] Remove charge prices code in "old arch" for confirmation/rejection/notification/persistence, so new arch may take over the responsibility
- [ ] Consider what to do with the code for publishing ChargePricesUpdated; it will probably be moved in story #1413
- [x] Remove Position column from the ChargePoint table. OBS: might cause some side effects in code base on outgoing flow as well. https://github.com/Energinet-DataHub/geh-charges/pull/1503

Other tasks written at PI planning, which may no longer be relevant/in sync with the code base:
[...]

PR #1503, opened 2022-07-27
feat: remove position from charge point

Mech0z:
## Description

When we need it, we calculate it instead

## References

* #1411


x-platformcoder (approved):
LGTM

PR #1507, opened 2022-08-01
Fix: TaxIndicatorMustBeFalse validation rules should not trigger when TaxIndicator is unknown

prtandrup:
## Description

When submitting a charge information request (D18) where taxIndicator is omitted from the request, like seen below, then it triggers the following validation rules, depending on the charge type; fee or subscription: 

* `TaxIndicatorMustBeFalseForSubscriptionValidationRule`
* `TaxIndicatorMustBeFalseForFeeValidationRule`

````
<!--cim:taxIndicator>false</cim:taxIndicator-->
````

This results in two errors in the provided rejection message:

````
        <cim:Reason>
            <cim:code>E0H</cim:code>
            <cim:text>Tax indicator must be set when calling with BusinessReasonCode D18 for charge type Subscription with charge ID SubId013 owned by 9656626091925.</cim:text>
        </cim:Reason>
        <cim:Reason>
            <cim:code>D14</cim:code>
            <cim:text>Tax indicator cannot be true for charge ID SubId013 owned by 9656626091925 as it is of charge type Subscription.</cim:text>
        </cim:Reason>
````

The second error may be confusing to the receiver, why this PR changes the mentioned validation rules, so they won't trigger upon receiving a request with omitted tax indicator. 
Rule `TaxIndicatorIsRequiredValidationRule` will still trigger and thus reject the request.

## References

* #0000


x-platformcoder (commented):
LGTM

x-platformcoder (approved):
OK

PR #1553, opened 2022-08-17
Fix/add price validation rules to new flow

jonasdmoeller:
## Description

Added validation rules for new price flow, so it now corresponds to the same level of validation in handling D08 messages in existing charge information flow.


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Dtos/Messages/Command/OperationBase.cs
```diff
@@ -13,17 +13,26 @@
 // limitations under the License.
 
 using GreenEnergyHub.Charges.Domain.Charges;
+using NodaTime;
 
 namespace GreenEnergyHub.Charges.Domain.Dtos.Messages.Command
 {
     public abstract class OperationBase
```
HenrikSommer, line 20: No longer a "Base" class?

HenrikSommer (approved):

Issue #1579, opened 2022-08-25
Bump Charges.Libraries nuget packages


prtandrup:
Maintenance task.

prtandrup:
### Regression test in T-001 

:warning: issue detected: correlationId problem between metering point and charges domains.

**Create Metering Point and verify default links:**
Metering Point ID: 577626689363303684, created.

Charges DB: 
![image.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/3706de83-1198-495e-a77f-f18bf8c01290)

Default charge links, Charges DB: Fails due to broken integration with Metering Point (Correlation-Id versus CorrelationID issue)
App insights [link](https://portal.azure.com/#view/AppInsightsExtension/DetailsV2Blade/ComponentId~/%7B%22SubscriptionId%22%3A%227b7ffd68-d27e-4c75-936a-2653ecc36450%22%2C%22ResourceGroup%22%3A%22rg-DataHub-SharedResouces-T-001%22%2C%22Name%22%3A%22appi-shared-sharedres-t-001%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252F7b7ffd68-d27e-4c75-936a-2653ecc36450%252FresourceGroups%252Frg-DataHub-SharedResouces-T-001%252Fproviders%252FMicrosoft.Insights%252Fcomponents%252Fappi-shared-sharedres-t-001%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel~/%7B%22eventId%22%3A%22b3c11914-2510-11ed-97dd-000d3aaba521%22%2C%22timestamp%22%3A%222022-08-26T07%3A27%3A10.947Z%22%2C%22cacheId%22%3A%22f48286cd-3da0-4cf6-b9f4-e63c0f3b58e1%22%2C%22eventTable%22%3A%22requests%22%2C%22timeContext%22%3A%7B%22durationMs%22%3A1800000%2C%22endTime%22%3A%222022-08-26T07%3A29%3A35.150Z%22%7D%7D)

prtandrup:
Kristian (Batman) has the task of fixing the issue in the Metering Point domain.

prtandrup:
Batman PR in the making: https://github.com/Energinet-DataHub/geh-metering-point/pull/920

prtandrup:
[...]

PR #1580, opened 2022-08-25
chore: Update Energinet.DataHub.MarketParticipant.Integration.Model

HenrikSommer:
## Description

The MarketParticipant domain has updated their `Energinet.DataHub.MarketParticipant.Integration.Model` NuGet package, and the production system is reliant on consuming domains to use the new package.

## References

* #1579


x-platformcoder (approved):
Approved

Issue #1289, opened 2022-04-21
Not possible to re-use a charge that was stopped on its' creation date.


prtandrup:
As a ChargeOwner
I want the ability to cancel a stop of a charge that was stopped on the charge's creation date
So that I may re-use a newly created (and by mistake) stopped charge.

**Notes:** 
- Cancel stop time limit apply.

**Original issue:**
A charge was created and then stopped (FeeId001) in T-001.
Currently it is not possible to update it with a new period.

ApplicationInsights [log](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22SubscriptionId%22%3A%227b7ffd68-d27e-4c75-936a-2653ecc36450%22%2C%22ResourceGroup%22%3A%22rg-DataHub-SharedResouces-T-001%22%2C%22Name%22%3A%22appi-shared-sharedres-t-001%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252F7b7ffd68-d27e-4c75-936a-2653ecc36450%252FresourceGroups%252Frg-DataHub-SharedResouces-T-001%252Fproviders%252FMicrosoft.Insights%252Fcomponents%252Fappi-shared-sharedres-t-001%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel/%7B%22eventId%22%3A%22e27561d9-c143-11ec-997c-2818787aa4fa%22%2C%22timestamp%22%3A%222022-04-21T07%3A22%3A20.995Z%22%2C%22cacheId%22%3A%22453adde1-bc84-49b4-8507-99b797d5e9d9%22%2C%22eventTable%22%3A%22requests%22%2C%22timeContext%22%3A%7B%22durationMs%22%3A1800000%2C%22endTime%22%3A%222022-04-21T07%3A24%3A43.369Z%22%7D%7D).

**Acceptance criteria:**
- [x] 1 Given a Charge (of any type) has been created and stopped on the same date
When the ChargeOwner does an update with effective date = created/stopped date
Then this cancels the stop and the charge is back "in use".

prtandrup:
### Tested in T-001 :heavy_check_mark:

Created new Subscription (SubId023):
[...]

PR #1594, opened 2022-08-30
Fix: Allow "CancelStop" on a charge that was created and subsequently stopped the same date

prtandrup:
## Description

This PR fixes the problem where a charge, that was created and stopped the same date, wouldn't allow a cancel stop operation.

Now...
* It allows a Charge to be persisted with a charge period where start and end date are equal.
* Upon cancelling the stop, it removes the charge period with matching start and end date.

## References

* #1289

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
@@ -208,6 +208,10 @@ namespace GreenEnergyHub.Charges.Domain.Charges
             var rules = GenerateRules(chargePeriod, taxIndicator, resolution, operationId).ToList();
             CheckRules(rules);
 
+            // Handles the case where a charge was created and subsequently stopped the same date
+            if (existingLastPeriod.StartDateTime == existingLastPeriod.EndDateTime)
+                _periods.Remove(existingLastPeriod);
+
             _periods.Add(chargePeriod);
         }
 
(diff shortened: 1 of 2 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Domain/Charges/Charge.cs
```diff
@@ -208,6 +208,10 @@ public void CancelStop(ChargePeriod chargePeriod, TaxIndicator taxIndicator, Res
             var rules = GenerateRules(chargePeriod, taxIndicator, resolution, operationId).ToList();
             CheckRules(rules);
 
+            // Handles the case where a charge was created and subsequently stopped the same date
+            if (existingLastPeriod.StartDateTime == existingLastPeriod.EndDateTime)
+                _periods.Remove(existingLastPeriod);
+
```
HenrikSommer, line 214: I generally prefer methods with describing names, rather than comments. Could we put this code in a RemovePeriodIfChargeWasCreatedAndSubsequentlyStoppedOnSameDate() method and remove the comment?

prtandrup: Sure, will do that.

prtandrup: done in https://github.com/Energinet-DataHub/geh-charges/pull/1594/commits/5632bf96ec78ed280b8f0b342850f08a3f65d575

HenrikSommer (approved):

PR #1596, opened 2022-08-31
Use B2C classes from Test Common

dstenroejl:
## Description

We have moved some reusable classes to Test Common.
In this PR we have refactored the code to use these classes.


HenrikSommer (commented):
Nice simplification in Charges

prtandrup (approved):
LGTM - and thanks for updating Charges to use the latest package 🏆

PR #1599, opened 2022-08-31
Spike using new JwtTokenValidator

dstenroejl:
## Description

Using if the refactored JwtTokenValidator. Should avoid retrieving OpenID configurations for every request. 

## References

https://github.com/Energinet-DataHub/geh-core/pull/156


HenrikSommer (approved):
LGTM, and approved on condition final release version is used for all packages.

HenrikSommer (approved):
Final versions in use 👌

Issue #1598, opened 2022-08-31
No default charge links assigned to new metering points.


PerTHenriksen:
See bug in metering point domain. 

Metering point domain receive a reply during meteing point creation, that CorrelationID is not set in charges domain, and thus no default charge links are assigned.

https://app.zenhub.com/workspaces/batman-60a6105157304f00119be86e/issues/energinet-datahub/geh-metering-point/923

prtandrup:
Our nuget packages `Energinet.DataHub.Charges.Clients` and `Energinet.DataHub.Charges.Clients.Registration` have been published in v. 3.0.10, where `OperationCorrelationId` is included, when a `CreateDefaultChargeLinksRequest` is sent to Charges (from Metering Point domain). Using those two new versions in the Metering Point domain requires a new pull request which Kristian F (Batman) takes care of. Pending this.

prtandrup:
### Tested in T-001 :heavy_check_mark:

Create new consumption metering point with ID: **571041421824913901**

NotifyBillingMasterData_MarketDocument (RSM-031) received: 

````
<?xml version="1.0" encoding="utf-8"?>
<cim:NotifyBillingMasterData_MarketDocument xmlns:cim="urn:ediel.org:structure:notifybillingmasterdata:0:1">
    <cim:mRID>826449d9b51e43f88d8eaf7fe7f16027</cim:mRID>
    <cim:type>D07</cim:type>
    <cim:process.processType>D17</cim:process.processType>
    <cim:businessSector.type>23</cim:businessSector.type>
    <cim:sender_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:sender_MarketParticipant.mRID>
    <cim:sender_MarketParticipant.marketRole.type>DDZ</cim:sender_MarketParticipant.marketRole.type>
    <cim:receiver_MarketParticipant.mRID codingScheme="A10">8100000000016</cim:receiver_MarketParticipant.mRID>
    <cim:receiver_MarketParticipant.marketRole.type>DDM</cim:receiver_MarketParticipant.marketRole.type>
    <cim:createdDateTime>2022-09-02T07:57:53Z</cim:createdDateTime>
    <cim:MktActivityRecord>
        <cim:mRID>a62c77b8b0414406b94643f3afe2db55</cim:mRID>
        <cim:marketEvaluationPoint.mRID codingScheme="A10">571041421824913901</cim:marketEvaluationPoint.mRID>
        <cim:ChargeGroup>
[...]

PR #1600, opened 2022-08-31
fix: `Energinet.DataHub.Charges.Client` - add custom service bus message property `OperationCorrelationId` to fix broken integration with Metering Point domain

HenrikSommer:
## Description

When the Metering Point domain sends a `Create default charge link request` to the Charges domain, it's done using Charges' NuGet package `Energinet.DataHub.Charges.Client`. With this PR, the package now adds a custom property called `OperationCorrelationId` to the service bus message being sent to Charges. This property was missing and caused the integration between the two domains to break.

Also, new `OperationCorrelationId` key set in `ApplicationProperties` of `ServiceBusMessage` in reply to a CreateDefaultChargeLinks request.

`Energinet.DataHub.Charges.Client` is bumped to v. 3.0.10

## References

* #1598

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/ReplySender/ServiceBusReplySender.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/ReplySender/ServiceBusReplySender.cs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Azure.Messaging.ServiceBus;
+using GreenEnergyHub.Charges.Infrastructure.Core.MessageMetaData;
 
 namespace GreenEnergyHub.Charges.Infrastructure.ReplySender
 {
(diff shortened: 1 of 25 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/ReplySender/ServiceBusReplySender.cs
```diff
@@ -29,11 +30,15 @@ public ServiceBusReplySender(ServiceBusSender serviceBusSender)
...
-            }).ConfigureAwait(false);
+            await _serviceBusSender.SendMessageAsync(
+                new ServiceBusMessage(data)
+                {
+                    CorrelationId = correlationId,
+                    ApplicationProperties =
+                    {
+                        new KeyValuePair<string, object>(MessageMetaDataConstants.CorrelationId, correlationId),
+                    },
+                }).ConfigureAwait(false);
```
prtandrup, line 41: Should we use the `ServiceBusMessageFactory.cs` here?

prtandrup (approved):
Approved with a single comment

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Infrastructure/ReplySender/ServiceBusReplySender.cs
HenrikSommer: Initially, I thought this was a good idea, but it turns out to be a larger change than first assumed - so maybe in a later PR.

PR #1619, opened 2022-09-06
New issue template for team Volt

PerLysemose:
Update issue template to align across teams


.github/ISSUE_TEMPLATE/story-template-volt.md
```diff
@@ -0,0 +1,21 @@
...
+### Acceptance Criteria
+________________
+### Tech. Notes
+________________
+### Task List
+_________________
+#### Tasks in Development
+_________________
+#### Tasks Completed
+_________________
```
prtandrup, line 21: How about just having one task list, where we can tick off completed tasks? This way we don't need to move stuff around (which probably will be forgotten any ways... 😄 )

```suggestion
### Task List
- [ ] <add task>
```

HenrikSommer (commented):
Please provide a pull request description

.github/ISSUE_TEMPLATE/story-template-volt.md
prtandrup: According to PerL, this solution has proven useful in Mandalorian; when issue is viewed directly in GitHub the task can be dragged and dropped. Let's give it a try otherwise we can change it back 😄 Resolved.

Issue #1589, opened 2022-08-29
Create new subscriptions for integration events and topics for internal events (New SB strategy 1/3)


Mech0z:
**Synopsis**
In the first part we create subscriptions on the new integrations-event topic, these subscriptions must be created in the Charges domain compared to our current subscriptions which are in shared domain. We also need to create new topics of all the existing topics currently in Charges service bus namespace in the shared service bus namespace.

We need to create these subscriptions using the component https://github.com/Energinet-DataHub/geh-terraform-modules/pull/96 and add SQL filters to limit to the correct files as all integration-events are collected in a single place.

**Existing internal charges topics**
![image.png](https://images.zenhubusercontent.com/60bdbd325b600c62f9576fbb/3380dd0a-5c42-44f7-ac63-45da0cf54b6f)

**Acceptance criteria**

1. - [ ] Subscription for Metering Point Created using correlation filter
2. - [ ] Subscription for Market Participant Changed using correlation filter
3. - [ ] Integration events Subscriptions are placed in Charges CI/CD
4. - [ ] internal events Subscriptions are placed in Charges CI/CD
5. - [ ] All individual internal SB topics and subscriptions are deleted in CI/CD
6. - [X] Existing internal events topics in Charges SB namespace have been copied to shared namespace
7. - [X] Charges provides all servicebus metadata required by GreenEnergyHub [ADR-0008](https://github.com/Energinet-DataHub/green-energy-hub/blob/a24e1c05b4b29f6b29963a23e7cea1fd0e94b3f0/docs/architecture-decision-record/ADR-0008%20Integration%20events.md)
8. - [ ] Run regression test to verify that path through the system still works

![image.png](https://images.zenhubusercontent.com/60bdbd325b600c62f9576fbb/09ce23fe-d22b-4a13-bd11-a9fe714d141e)

PerLysemose:
@PerTHenriksen I have closed this issue. There is nothing to demo on it. Verification that the story is complete is shown in the picture above. Please reach out to Jonas in case of any questions.

PR #1631, opened 2022-09-09
feat: Add message metadata to ServiceBus messages as required by GreenEnergyHub ADR-0008

HenrikSommer:
## Description

This pull request isn't a core part of the solution to #1589, but a key part for other domains to solve a similar story:
We add all message metadata mentioned in [ADR-0008](https://github.com/Energinet-DataHub/green-energy-hub/blob/a24e1c05b4b29f6b29963a23e7cea1fd0e94b3f0/docs/architecture-decision-record/ADR-0008%20Integration%20events.md)

## References

* #1589


Mech0z:
Think we should add metadata to our internal events as well, but could be in a separate PR

Mech0z (approved):

Issue #1562, opened 2022-08-17
Outbox v2 - 4. Introduce service bus event for raising integration events


HenrikSommer:
![OutboxV2_4.jpg](https://images.zenhubusercontent.com/60bf731f5b600c8c1257787b/c6dd64b0-ecb2-46cf-994f-9ab522d4d561)

**Synopsis**
As Charges domain with an outbox solution implemented,
I want to publish an integration event when charge changed outbox message is persisted,
So that other domains within DataHub can subscribe to events from the Charges domain with a high level of reliability.

**Acceptance criteria**

- [x] 1. Given a charge price update
When an confirmation event is raised on internal service bus,
Then an integration event is published on a shared service bus
<br>

prtandrup:
Blocked due to MessageHub issue in T-001

prtandrup:
### Tested in T-001 :heavy_check_mark:

Submitted Charge Price (D08) to a tax tariff, id: SyoTar0004:
Request:
````
<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
    <cim:mRID>DocId2022-09-15T06:11:40.084Z</cim:mRID>
    <cim:type>D10</cim:type>
    <cim:process.processType>D08</cim:process.processType>
    <cim:businessSector.type>23</cim:businessSector.type>
    <cim:sender_MarketParticipant.mRID codingScheme="A10">5790000432752</cim:sender_MarketParticipant.mRID>
    <cim:sender_MarketParticipant.marketRole.type>EZ</cim:sender_MarketParticipant.marketRole.type>
    <cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
    <cim:receiver_MarketParticipant.marketRole.type>DDZ</cim:receiver_MarketParticipant.marketRole.type>
    <cim:createdDateTime>2022-09-15T06:11:40Z</cim:createdDateTime>
    <cim:MktActivityRecord>
        <cim:mRID>OpId2022-09-15T06:11:40.084Z</cim:mRID>
[...]

PR #1639, opened 2022-09-13
feat: split chargeprice integration event from old flow into new chargeprice flow

x-platformcoder:
## Description
The charge  price flow will send integration events to other domains when confirmed/accepted chargeprices (D08) has been ingested, handled and persisted

## References

* #1562

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/Factories/ChargePricesUpdatedEventFactory.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/Factories/ChargePricesUpdatedEventFactory.cs
@@ -13,21 +13,21 @@
 // limitations under the License.
 
 using GreenEnergyHub.Charges.Application.Charges.Acknowledgement;
-using GreenEnergyHub.Charges.Domain.Dtos.ChargeInformationCommands;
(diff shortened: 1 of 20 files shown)


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Application/Charges/Factories/ChargePricesUpdatedEventFactory.cs
```diff
@@ -13,21 +13,21 @@
...
-                chargeInformationOperationDto.ChargeType,
-                chargeInformationOperationDto.ChargeOwner,
-                chargeInformationOperationDto.StartDateTime,
-                chargeInformationOperationDto.EndDateTime.GetValueOrDefault(),
-                chargeInformationOperationDto.Points);
+                chargePriceOperationDto.SenderProvidedChargeId,
+                chargePriceOperationDto.ChargeType,
+                chargePriceOperationDto.ChargeOwner,
+                chargePriceOperationDto.StartDateTime,
+                chargePriceOperationDto.EndDateTime.GetValueOrDefault(),
```
prtandrup, line 29: I believe we should switch to these two instead: 
```suggestion
                chargePriceOperationDto.pointsStartInterval,
                chargePriceOperationDto.pointsEndInterval,
```

prtandrup (approved):
Approved with a single comment.

Issue #1588, opened 2022-08-29
Additional async validation rules need to be applied to new price flow (D08)


prtandrup:
Additional asynchronous validation rules have been identified as missing in the new price flow (for Charge Prices D08 processing)

The following rules must be applied in the new price flow:

* ResolutionValidationRule (to apply for subscription, fee and tariff price series) (VR 505-x )
* StartDateValidationRule (applied on effectiveDate of price series) (VR 209)
* PreviousOperationsMustBeValidRule (already applied, but the namespace seems to be charge information relevant only)
* ResolutionIsRequired (VR912)

**Acceptance criteria**
- [x] 1 - Above mentioned asynchronous validation rules have been applied to D08 Charge Prices requests.
- [x] 2 - The asynchronous validation rules md [file](https://github.com/Energinet-DataHub/geh-charges/blob/main/docs/business-processes/asynchronous-validations.md) has been updated to match this.

prtandrup:
Blocked due to MessageHub issue in T-001

x-platformcoder:
**Ingestion**
````
<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
    <cim:mRID>DocId{{$isoTimestamp}}</cim:mRID>
    <cim:type>D10</cim:type>
    <cim:process.processType>D08</cim:process.processType>
    <cim:businessSector.type>23</cim:businessSector.type>
    <cim:sender_MarketParticipant.mRID codingScheme="A10">{{MarketParticipantGln}}</cim:sender_MarketParticipant.mRID>
    <cim:sender_MarketParticipant.marketRole.type>{{MarketParticipantRole}}</cim:sender_MarketParticipant.marketRole.type>
    <cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
    <cim:receiver_MarketParticipant.marketRole.type>DDZ</cim:receiver_MarketParticipant.marketRole.type>
    <cim:createdDateTime>{{ISO8601Timestamp}}</cim:createdDateTime>
[...]

PR #1658, opened 2022-09-15
Fix StartValidationRule error text for ChargePrice

ShaBloze:
## Description

When StartValidationRule fails, the ChargeStartDateTime is filled with unknown.

## References

* #1588


x-platformcoder (approved):
LGTM

Issue #1651, opened 2022-09-15
Ensure integration test, code, and infrastructure-as-code relationship


HenrikSommer:
**Synopsis**
As a developer, I want to be able to refactor the charge domain code base, without the risk of compromising the relationship between code and infrastructure-as-code.

**Tech notes**
* ServiceBus subscription filters for charge domain events are tightly coupled to inheritances of the _DomainEvent_
* _EnvironmentSettingNames_ are loosely coupled to _ChargesServiceBusResourceNames_ but it would be nice if they do not diverge/separate

These are the files involved:
* _EnvironmentSettingNames_
* _ChargesServiceBusResourceNames_
* _func-functionhost.tf_
* _sbt-charges-domain-events.tf_
* _local.settings.sample.json_
* _DomainEvent_ (and its inheritances)

**Acceptance criteria:**

- [x] 1. Given _DomainEvent_,
           when any inherited class' name is changed
           then it must be unambiguous to the developer, that ServiceBus subscription filters need to be changed in accordance

- [ ] ~~2. Given _EnvironmentSettingNames_,
           when any setting name is changed
           then it must be unambiguous to the developer, that _func-functionhost.tf_ needs to be changed in accordance~~

- [x] 3. Given _EnvironmentSettingNames_,
           when any setting name is changed
           then it must be unambiguous to the developer, that _ChargesServiceBusResourceNames_ need to be changed in accordance

- [x] 4. Given _ChargesServiceBusResourceNames_,
           when any setting name is changed
           then it must be unambiguous to the developer, that _EnvironmentSettingNames_ need to be changed in accordance

- [ ] ~~5. Given _DomainEvent_,
           when any inherited class is renamed without changing infrastructure-as-code
           then healthCheck should fail~~

- [ ] ~~6. Nice-to-have, but probably not worth it:
           Given _func-functionhost.tf_,
           when any domain event related setting is changed
           then it must be unambiguous to the developer, that _ChargesServiceBusResourceNames_ need to be changed in accordance~~

HenrikSommer:
Tests directly including infrastructure-as-code are skipped, due to our expectation that it will be taken private. The extension of health checks to include correlation filters is postponed but could be reconsidered when infrastructure-as-code is taken private

HenrikSommer:
Testing: This is a purely technical story, and the solution has been presented to team members. One improvement was made as a result of the presentation.

PR #1660, opened 2022-09-16
test: 1651 Ensure integration test code and infrastructure as code relationship

HenrikSommer:
## Description

As a developer, I want to be able to refactor the charge domain code base, without the risk of compromising the relationship between code and infrastructure-as-code. This Pull request does not ensure an exact relationship but it does go quite a long way:
* [X] secure relationship between `DomainEvent` inheritances name and the correlation filter, that is presumed to be used on the corresponding service bus subscription. With this, the developer will be reminded to also rename, create or delete a subscription in Charges infrastructure-as-code
* [X] secure relationship between `EnvironmentSettings` used in Charges `FunctionHost` and `ChargeServiceBusResourceNames` used in Charges domain/integration tests. With this, the developer will be reminded to also rename, create or delete an `EnvironmentSetting` or `ChargeServiceBusResourceName` when the other part changes

**Tech notes**
* Before long, Infrastructure-as-Code is presumably not going to be part of Charges domain, so we have not put any effort into testing the exact relationship with concrete Terraform files
* Apparently, it is not possible to do ServiceBus subscriptoin HealthChecks with check of a specific correlation filter
* Test data builders are moved to the `TestCore` project to avoid reference from `IntegrationTests` to `Tests`. The downside of this is that the `TestCore` project now references the `Domain`, `Application`, and `Infrastructure` projects, and the last part creates a circular reference in the `Libraries` solution, which is solved by removing the usage of `TestCore` in `Libraries` solution. We could consider a `TestData` project, to put the test data builders in, thereby removing the necessity of the `Infrastructure` reference that created the circular reference

## References

* #1651


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/FunctionHost/EnvironmentSettingNamesTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            var domainEventChargesServiceBusResourceNames =
+                DomainEventSettingHelper.GetDomainEventSettings(typeof(ChargesServiceBusResourceNames));
+            var settings = domainEventEnvironmentSettingNames
+                .Select(domainEventEnvironmentSettingName =>
+                    Environment.GetEnvironmentVariable(domainEventEnvironmentSettingName.Value));
+
+            // Act
+            // Assert
+            foreach (var setting in settings)
+                domainEventChargesServiceBusResourceNames.Values.Should().Contain(setting);
```
jonasdmoeller, line 50:
```suggestion
            foreach (var setting in settings)
            {
                domainEventChargesServiceBusResourceNames.Values.Should().Contain(setting);
            }
```

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Infrastructure.Core/InternalMessaging/ChargeServiceBusResourceNamesTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            var domainEventChargesServiceBusResourceNames =
+                DomainEventSettingHelper.GetDomainEventSettings(typeof(ChargesServiceBusResourceNames));
+            var settings = domainEventSettings
+                .Select(domainEventEnvironmentSettingName =>
+                    Environment.GetEnvironmentVariable(domainEventEnvironmentSettingName.Value)).ToList();
+
+            // Act
+            // Assert
+            foreach (var domainEventChargesServiceBusResourceName in domainEventChargesServiceBusResourceNames)
+                settings.Should().Contain(domainEventChargesServiceBusResourceName.Value);
```
jonasdmoeller, line 49:
```suggestion
            foreach (var domainEventChargesServiceBusResourceName in domainEventChargesServiceBusResourceNames)
            {
                settings.Should().Contain(domainEventChargesServiceBusResourceName.Value);
            }
```

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/Startup.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace GreenEnergyHub.Charges.Tests
+{
+    public class Startup
+    {
+        [UsedImplicitly]
+        public static void ConfigureHost(IHostBuilder hostBuilder)
+        {
+            var config = new ConfigurationBuilder()
+                .AddJsonFile("local.settings.sample.json")
```
jonasdmoeller, line 27: Shouldn't this be the actual local settings file instead of the sample file?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/TestHelpers/DomainEventTypeHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// </summary>
+        public static IEnumerable<string> GetOrderedDomainEventTypeNames()
+        {
+            var domainAssembly = AssemblyHelper.GetDomainAssembly();
+            var messageTypes = domainAssembly
+                .GetTypes()
+                .Where(t => typeof(DomainEvent).IsAssignableFrom(t) && t.IsClass && !t.IsAbstract)
+                .Select(t => t.Name)
+                .ToList();
+            return messageTypes.OrderBy(x => x).ToList();
```
jonasdmoeller, line 35: Why call ToList() twice?

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/TestHelpers/PresumedServiceBusFilterHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+                "ChargePriceCommandReceived" + "Event",
+                "ChargeLinksRejected" + "Event",
+                "ChargeLinksReceived" + "Event",
+                "ChargeLinksDataAvailableNotified" + "Event",
+                "ChargeLinksAccepted" + "Event",
+                "ChargeInformationCommandReceived" + "Event",
+                "ChargeInformationCommandRejected" + "Event",
+                "ChargeInformationCommandAccepted" + "Event",
+            };
+            return presumedServiceBusFilters.OrderBy(x => x).ToList();
```
jonasdmoeller, line 42: Is .ToList() necessary?

jonasdmoeller (changes requested):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/TestHelpers/DomainEventTypeHelper.cs
HenrikSommer: Will remove the first :-)

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/TestHelpers/PresumedServiceBusFilterHelper.cs
HenrikSommer: Actually not, will remove and change method signature to return `IOrderedEnumerable<string>`

jonasdmoeller (approved):

Issue #1328, opened 2022-05-02
Receive and process charge prices in chronological order (Feature-AC11a)


prtandrup:
**Synopsis:** 
As a Grid Access Provider or System Operator
I want to bundle multiple charge price operations (D08) for the same charge price in a single charge price request (other charges may be included as well)
So that I can minimize the number of requests I need to send

**Tech notes:**
* Implement regression test (integration test or domain test) verifying chronological order

**Acceptance criteria:** 
- [x] (1) Handle multiple price series, received in the same bundle for the same charge, in chronological order

- [x] (2) Multiple price series, received in the same bundle for the same charge, are communicated through MessageHub in chronological order

~~- [ ] (3) Multiple price series, received in the same bundle for the same charge, are communicated through integration events in chronological order~~

Mech0z:
Deemed that we are already have unittest covering this with CreateAsync_WhenSeveralOperationsInChargePriceCommand_ReturnOrderedListOfOperations for incoming and outgoing with GetAsync_WhenChargeDataAvailableIsOperationOrdered_ReturnsOrderedData

Integration test is limited by MessageHubSimulation not supporting bundling

We could create system test to validate this, but currently system test is not minded for this kind of testing but more 1 happy flow per way through the system.

HenrikSommer:
GetAsync_WhenChargeDataAvailableIsOperationOrdered_ReturnsOrderedData only works on ChargeInformation available data, so I suggest we at least extend this test or create a similar for ChargePrices. I will also look into the bundling issue.

x-platformcoder:
Checking chronological order for D18 in Integrationtests - OK
Checking chronological order for D18 in Postman - OK
 
```<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
    <cim:mRID>DocId{{$isoTimestamp}}</cim:mRID>
    <cim:type>D10</cim:type>
    <cim:process.processType>D18</cim:process.processType>
    <cim:businessSector.type>23</cim:businessSector.type>
[...]

PR #1671, opened 2022-09-21
test: Receive and process charge prices in chronological order - part I

HenrikSommer:
## Description

As for charge information we want to be sure that we create charge price notifications in the same order as they were received. For a start, we upgrade an existing test to ensure correct ordering in available data for all inheritances of `AvailableDataBase`.

Expectantly we will, in an upcoming pull request covering the same story, seek to ensure chronological ordering for both charge information and charge prices.

## References

* #1328


x-platformcoder (approved):
LGTM

PR #1678, opened 2022-09-23
CI optimization

dstenroejl:
## Description

Use new workflows so we build on Ubuntu and run tests on Windows. Further we can then split tests and have multiple runs in parallel.

## References

https://app.zenhub.com/workspaces/mighty-ducks---the-outlaws-6193fe815d79fc0011e741b1/issues/energinet-datahub/the-outlaws/585

PR #1697, opened 2022-09-26
Test: Change embedded test files to Content 

prtandrup:
## Description

As a prerequisite to the CI optimization work in this [PR](https://github.com/Energinet-DataHub/geh-charges/pull/1678) the test files that were previously added as embedded file are now changed to content (copy if newer).

As the CI will built on linux and do test execution test on windows, embedded xml files receive extra carriage return `\r` during the build process, which makes our comparison tests fail. Hence, this PR changes the embedded files to content, thus avoiding the carriage return problem

## References

* #0000


Mech0z (approved):
Looks good, but should we implement a concept like GreenEnergyHub.Charges.IntegrationTest.Core.TestFiles.Charges.ChargeDocument for unittests to avoid "magic strings"

PR #1707, opened 2022-09-30
refac: make a cleaner way to avoid problems with summer winter time

Mech0z:
## Description

This tries to make a cleaner way to avoid hitting the dates where we have either 23 or 25 hours.

## References

* #1706


x-platformcoder (approved):
LGTM

PR #1773, opened 2022-10-18
Fix: Price series ordered ascending in JSON notifications (based on position value)

prtandrup:
## Description

With this PR, a JSON charge price notification (D08) will now list price series in ascending order based on the position value.

Also, a shuffling of price test data was added to an existing test, to verify that this ordering actually works.

## References

* #0000

### DIFF
--- a/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Infrastructure/Cim/Bundles/Charges/ChargePriceCimJsonSerializerTests.cs
+++ b/source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Infrastructure/Cim/Bundles/Charges/ChargePriceCimJsonSerializerTests.cs
@@ -181,7 +181,7 @@ namespace GreenEnergyHub.Charges.Tests.MessageHub.Infrastructure.Cim.Bundles.Cha
             }
 
             // shuffle points order
-            var shuffledPoints = points.OrderBy(p => _rnd.Next()).ToList();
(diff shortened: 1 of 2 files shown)


Mech0z (approved):

source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.Tests/MessageHub/Infrastructure/Cim/Bundles/Charges/ChargePriceCimJsonSerializerTests.cs
```diff
@@ -175,12 +175,15 @@ private static List<AvailableChargePriceDataPoint> GetPoints(int noOfPoints)
...
 
-            for (int i = 1; i <= noOfPoints; i++)
+            for (var i = 1; i <= noOfPoints; i++)
             {
                 points.Add(new AvailableChargePriceDataPoint(i, i));
             }
 
-            return points;
+            // shuffle points order
+            var shuffledPoints = points.OrderBy(_ => _rnd.Next()).ToList();
```
HenrikSommer, line 184: In principle, a test using these randomly ordered points could sometimes result in an ordering that makes the test pass even if the production code did not order the points :-)

prtandrup: Especially for those that only has 1 point ;-) but I'll take my chances with the price series that holds 24 points :smile:

Issue #1763, opened 2022-10-14
RSM-033 RequestChangeOfPriceList D08 - Position doesn't match the indication of start and end, but the message is confirm instead of rejectingf a reject


IreneDataHub:
When DDM/Grid send a request RequestChangeOfPriceList/ RSM-033 D08 they get an confirm, but we expect a rejecting when position doesn't match timeperiod in the RequestChangeOfPriceList D08

Expect a reject 
![image.png](https://images.zenhubusercontent.com/612f6568004bdd03eab7ca46/ae37de20-0298-4ff9-8a6d-b79113a57d4f)

Please also investigate ticket #1762 - Maybe they are related
RSM-033 request RequestChangeOfPriceList D08 

<?xml version="1.0" encoding="UTF-8"?>
<cim:RequestChangeOfPriceList_MarketDocument xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cim="urn:ediel.org:structure:requestchangeofpricelist:0:1" xsi:schemaLocation="urn:ediel.org:structure:requestchangeofpricelist:0:1 urn-ediel-org-structure-requestchangeofpricelist-0-1.xsd">
	<cim:mRID>DocId{{$isoTimestamp}}</cim:mRID>
	<cim:type>D10</cim:type>
	<cim:process.processType>D08</cim:process.processType>
	<cim:businessSector.type>23</cim:businessSector.type>
	<cim:sender_MarketParticipant.mRID codingScheme="A10">{{MarketParticipantGln}}</cim:sender_MarketParticipant.mRID>
	<cim:sender_MarketParticipant.marketRole.type>{{MarketParticipantRole}}</cim:sender_MarketParticipant.marketRole.type>
	<cim:receiver_MarketParticipant.mRID codingScheme="A10">5790001330552</cim:receiver_MarketParticipant.mRID>
	<cim:receiver_MarketParticipant.marketRole.type>DDZ</cim:receiver_MarketParticipant.marketRole.type>
	<cim:createdDateTime>{{ISO8601Timestamp}}</cim:createdDateTime>
	<cim:MktActivityRecord>
		<cim:mRID>OpId{{$isoTimestamp}}</cim:mRID>
		<cim:ChargeGroup>
			<cim:ChargeType>
				<cim:chargeTypeOwner_MarketParticipant.mRID codingScheme="A10">{{MarketParticipantGln}}</cim:chargeTypeOwner_MarketParticipant.mRID>
[...]

PR #1781, opened 2022-10-19
fix: update how to calculate number of month

Mech0z:
## Description

this fixes a bug where calculation of month used UTC time, now it uses locallized time instead

If we used utc on 2022-06-14T22:00Z and 2022-07-31T22:00Z and used .month an subtracted them from each other we got 7-6=1, if its converted to localtime 2022-06-15 and 2022-08-01, the correct 8-6=2 month is calculated.

This also fixes an error in expected points in a test case with irregular price series going from summer to wintertime .

## References

* #1763


Mech0z:
Replaced by https://github.com/Energinet-DataHub/geh-charges/pull/1781

Issue #1231, opened 2022-03-25
Charge update must not have bound end date


Mech0z:
Instead of throwing an exception we should fail using a validation rule instead as the end user do not currently get a response when it fails.

**### Find out if this is still a problem, code have been refactored a lot since, if so create a validation rule instead of throwing an exception.**

Log https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22SubscriptionId%22%3A%223c13ff3d-aad8-4ee8-94c6-8a938ace9699%22%2C%22ResourceGroup%22%3A%22rg-DataHub-SharedResouces-U-001%22%2C%22Name%22%3A%22appi-shared-sharedres-u-001%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252F3c13ff3d-aad8-4ee8-94c6-8a938ace9699%252FresourceGroups%252Frg-DataHub-SharedResouces-U-001%252Fproviders%252FMicrosoft.Insights%252Fcomponents%252Fappi-shared-sharedres-u-001%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel/%7B%22eventId%22%3A%220241b406-ac13-11ec-997c-2818787a4d67%22%2C%22timestamp%22%3A%222022-03-25T08%3A09%3A33.788Z%22%2C%22cacheId%22%3A%227cbdf70f-691d-4338-9738-a820e03d6330%22%2C%22eventTable%22%3A%22exceptions%22%2C%22timeContext%22%3A%7B%22durationMs%22%3A86400000%2C%22endTime%22%3A%222022-03-25T12%3A21%3A15.064Z%22%7D%7D

GreenEnergyHub.Charges.Domain.Charges.Charges.cs
```
        public void Update(ChargePeriod newChargePeriod, TaxIndicator taxIndicator, Resolution resolution, string operationId)
        {
            ArgumentNullException.ThrowIfNull(newChargePeriod);
            ArgumentNullException.ThrowIfNull(operationId);

            // This should be in a separate rule and handled by ChargeOperationFailedException
            // in order to notify sender that operation failed because of this constraint
            if (newChargePeriod.EndDateTime != InstantExtensions.GetEndDefault())
            {
                throw new InvalidOperationException("Charge update must not have bound end date.");
            }
```

[...]

PR #1794, opened 2022-10-20
Fix: Remove irrelevant `newChargePeriod.EndDateTime` check in `Charge.Update()`: true cannot be reached

prtandrup:
## Description

This PR removes this code section in `Charge.Update()`:

image

Code cannot be reached due to validation rule VR.917, which guards any requests where `termination date` does not equal `effective date`. Without this rule it would have been possible to reach Charge.Update with an newChargePeriod.EndDateTime that was not equal `EndDefault`;

image

Right now, for an operation deemed `charge update`, the new charge period created will be set to `EndDefault` as `chargeInformationOperationDto.EndDateTime` is null: 

image

## References

* #1231


HenrikSommer (commented):
PR description is 🚀

HenrikSommer (approved):

Issue #1689, opened 2022-09-23
GUI - Add table view to messages tab in drawer


ManBearTM:
We would like to show a table of messages associated with a charge.

![messages-table.png](https://images.zenhubusercontent.com/626fbc6d95e6587091961d95/cc21d2a4-8f77-4d1c-a0d0-1e92dcab26c9
)

**Technical Notes**
Re-use the same component as in the prices tab.

How data is created and served:
![ChargeMessageReference.png](https://images.zenhubusercontent.com/60ae42d0916bb92d8b000901/c7cb12de-aef5-4d08-a80a-86e7da55dccb)

**Tasks**
- [x] Fetch data automatically (or just take from response, see below)
- [X] Display table with data
- [x] Hook up table data with date range picker

**Prerequisite tasks**
- [x] Table with repository for persisting message id (document Id), with reference to the charge (see ChargeMessageReference in diagram)
- [x] Two new functions in the Charges information and prices command flows: Reacts to ChargeInformation/Prices accepted events and persists message (document) id, with reference to the charge
- [x] Update the architectural diagram with the two new functions
- [x] Charges WebApi: New method to get messages for a charge
- [x] Charges.Clients: New method to get messages for a charge from Charges.WebApi
- [x] DataHub.WebApi (BFF): Update new Charges.Clients package and expose method to get messages for a charge

**Clarifications**
Is the data loaded asynchronously or included in the initial search response?

**Acceptance Criteria**
- [ ] As a GUI user, I want to be able to see a list of messages associated with a charge.

IreneDataHub:
Test T- environment

Message D18 and D08 is shows as expect for charge Irene Okt. 
Have also made som others test and it works as acc. said 
Languge is also ok in danish and english

![image.png](https://images.zenhubusercontent.com/612f6568004bdd03eab7ca46/b5d9b360-134e-4aa2-97cc-d5755e5b68c6)

PR #1795, opened 2022-10-20
feat: ChargeMessages added to command stack

HenrikSommer:
## Description

This is step one in the list of prerequisite tasks on issue #1689:
- [X] Table with repository for persisting message id (document Id), with reference to the charge

## References

* #1689


prtandrup (approved):
Approved, but please consider if the fulfillment of DATO and TYPE data needs must be accounted for in this PR as well? Think we forgot about it, when we talked yesterday.

image

HenrikSommer (commented):
Thank you for the review, @prtandrup. I don't think we need to register time or type. Those data are saved together with the message in the message archive.

PR #1805, opened 2022-10-24
refac: Environment variables in appsettings

ShaBloze:
## Description

Right now we have to put environment variables into `launchSettings.json` to start our WebApi local. We would like to have those variables in the appsettings.

## References

* #0000


x-platformcoder (approved):
LGTM

PR #1815, opened 2022-10-26
feat: 1689 webapi get charge messages for charge

HenrikSommer:
## Description

Charges.WebApi is extended with a ChargeMessage search endpoint.

During the implementation of the endpoint, it became clear that we can provide better search and sorting capabilities for the DataHub frontend with message type and date added, which is included here.

Not all test cases are completed. This will be done in an upcoming pull request.

## References

* #1689


ShaBloze (approved):
LGTM

PR #1824, opened 2022-10-31
feat: Return ChargeMessagesV1Dto with total count and content

HenrikSommer:
## Description

To enable pagination in the frontend view of messages, we return an object with total count and the relevant messages.

## References

* #1689


source/GreenEnergyHub.Charges/source/GreenEnergyHub.Charges.QueryApi/QueryServices/ChargeMessageQueryService.cs
```diff
@@ -71,9 +75,7 @@ public async Task<IEnumerable<ChargeMessageV1Dto>> SearchAsync(ChargeMessagesSea
                              cm.Type == charge.Type &&
                              cm.MarketParticipantId == marketParticipant.MarketParticipantId)
                 .Where(cm => cm.MessageDateTime >= searchCriteria.FromDateTime &&
-                             cm.MessageDateTime < searchCriteria.ToDateTime)
-                .Skip(searchCriteria.Skip)
-                .Take(searchCriteria.Take);
+                             cm.MessageDateTime < searchCriteria.ToDateTime);
```
ShaBloze, line 78: On ChargePrices we changed it to include the `ToDateTime`, so I think we should do the same here.

ShaBloze (approved):
LGTM with one comment.

PR #1827, opened 2022-11-01
Move infrastructure to private repository

djorgensendk:


.github/workflows/cd.yml
```diff
@@ -72,19 +61,14 @@ jobs:
...
     ]
     if: |
       always() &&
       !contains(needs.*.result, 'failure') &&
       !contains(needs.*.result, 'cancelled') &&
-      !(
-        needs.infrastructure_promote_prerelease.result == 'skipped' &&
-        needs.dotnet_promote_prerelease.result == 'skipped'
-      )
+      needs.dotnet_promote_prerelease.result == 'skipped'
```
dstenroejl, line 71: I think the expression has now been inverted. Maybe we should just keep the format we have "!( xxx )", like we did in other domains even though there was only 1 line. This makes it less error prone when changing the expression.

dstenroejl (changes requested):

.github/workflows/cd.yml
djorgensendk: changed :)

dstenroejl (approved):

dstenroejl (approved):

PR #1847, opened 2022-11-03
Msp/serilog test

Mech0z:
## References

* #0000


HenrikSommer (commented):
We should close this PR, now that it is replaced by #1864 with `Microsoft.Azure.Functions.Worker.ApplicationInsights`

Issue #1846, opened 2022-11-03
Persist Charge Information History and enable the Web API to query the data


prtandrup:
Based on the persistence technology selected in #1841, this story is about implementing the historical data model for charge information and being able to populate it with data from accepted Charge Information events, so it can be queried through Charges Web API.

Decision from #1841: Built a custom solution based on service bus, a history function and custom history table(s).

This is considered our query stack.

**Acceptance criteria:**

- [x] 1 Documentation created/updated in GitHub:
* **Charge Information History data model** to be documented [here](https://github.com/Energinet-DataHub/geh-charges/tree/main/docs/data-models).
* ~~**Charge Information History persistence flow** to be documented, similar to [this](https://github.com/Energinet-DataHub/geh-charges/tree/main/docs/process-flows#persist-charge-information).~~
* **Architecture diagram** updated

- [x] 2 Given a "charge history" Web API request
When inspecting the response 
The response must as a minimum include the following data:

* Period start
* Period end
* Name
* Description
* Resolution
* Transparent invoicing
* VAT classification
* Tax
* Type
* Owner
* (Additional data might be needed for coupling it with the charge)

- [x] 3 Given a charge has been created with the following periods (simplified to name updates only)

* 01/01/2022 08:02 AM: ChargeA is created with Name: "Name A0" with effective date: 01/02/2022
* 01/01/2022 08:15 AM: ChargeA is updated with Name: "Name A1" with effective date: 01/02/2022
* 03/01/2022 10:00 PM: ChargeA is updated with Name: "Name B" with effective date: 01/04/2022
* 04/01/2022 09:00 PM: ChargeA is updated with Name: "Name A2" with effective date: 01/02/2022
* 05/01/2022 05:01 AM: ChargeA is updated with Name: "Name future" with effective date: 01/01/2024

When below requests are made, 
Then the following response are returned

[**a**] Request with datetime input: 01/01/2022 08:14 AM 
the response contains 
ChargeA with "Name A0" per 01/02/2022

[**b**] Request with datetime input: 01/01/2022 08:16 AM 
the response contains 
ChargeA with "Name A1" per 01/02/2022

[**c**] Request with datetime input: 03/01/2022 10:01 PM:
the response contains 
ChargeA with "Name A1" per 01/02/2022
ChargeA with "Name B" per 01/04/2022 

[**d**] Request with datetime input: 04/01/2022 09:01 PM:
the response contains 
[...]

PR #1871, opened 2022-11-17
Docs: Add Charge History data model

prtandrup:
## Description

This PR adds the Charge History data model documentation.

Decisions: 

* Keep table in NF1
* `Charge.Id (Guid)` and `MarketParticipant.Id (Guid)` will not be persisted, the combined unique key `SenderProvidedChargeId + Type + Owner` provided in the `ChargeInformationAcceptedEvent` should be enough when querying the data from Frontend.

## References

* #1846


x-platformcoder (approved):
fine

